PR #11, opened 2024-10-07
fix: BootNotification must respect the OCPP specification

juherr:
The issue comes from `request.BootNotificationRequest` which is an empty object and then tsc accept everything instead of "empty object only".


shiv3:
@juherr thank you for taking PR!! seems great 👍👍  I will confirm this in today.

shiv3:
I've confirmed that it works, so I'll merge it. Thank you for the fix PR.

PR #12, opened 2024-10-08
Clean transaction when the startTransaction message is refused

juherr:


src/cp/OCPPMessageHandler.ts
```diff
@@ -421,6 +417,13 @@ export class OCPPMessageHandler {
       this._logger.log("Failed to start transaction");
       if (connector) {
         connector.status = OCPPStatus.Faulted;
+        if (connector.transaction && connector.transaction.meterSent) {
+          this._chargePoint.stopTransaction(connectorId);
+        } else {
+          this._chargePoint.cleanTransaction(connectorId);
+        }
+      } else {
+        this._chargePoint.cleanTransaction(connectorId);
```
shiv3, line 426: I think it's a good idea to stop the transaction if it already exists, and clean it if it doesn't. LGTM 👍

shiv3 (approved):
These fixes are great, thank you!

src/cp/OCPPMessageHandler.ts
juherr: In fact, I think it is the best behavior by default but it should be configurable too  because it can simulate a bad charger worflow and help to check how the csms is working in that situations.

PR #13, opened 2024-10-09
Fix: reset meterSent during transaction cleanup

juherr:
make connector available at the end of the transaction
refacto: avoid connector fetching when not needed


juherr:
I'm not really sure about the refacto and I can rollback it if needed.

shiv3 (approved):
lgtm

PR #17, opened 2024-10-17
Implement hard/soft reset command

juherr:


shiv3 (approved):
nice, I feel name of  reset and boot is good

PR #18, opened 2024-10-17
fix: the result payload must not have action

juherr:
According to the specification: 
image


shiv3 (approved):
I see, thank you for refactoring 👍

PR #26, opened 2025-07-21
feat: desktop application that Tauri based 

shiv3:
occp-cp-simulator was a browser-based standalone page application, 
but since it seems difficult to solve issues such as CORS with a web page-based application, 
and version control seems difficult with GithubPages,
so I will try to make it into a desktop application based on Tauri.

image


juherr:
@shiv3 The idea is great! However, try to keep it browser-compatible if possible: there’s no CORS issue when using it with unsecured WebSockets or when it’s directly integrated into a CSMS.

shiv3:
@juherr 
Thank you! Yes, I plan to continue using it as a browser-based browser and maintain compatibility, but my biggest motivation was wanting to break through the upper limit on the number of Websocket connections that Chromium-based browsers had. There are still few functions and I think the installation is troublesome, so I think I will continue to use it as a browser-based browser.

Issue #101, opened 2026-06-30
Scenario editor can not save the new upload scenario after refreshing the Web


CHENJIANMIN-git:
### Test step
1. upload a new scenario file
2. save
3. refresh the Web
### Result
The scenario is still the old one.

<img width="1623" height="853" alt="Image" src="https://github.com/user-attachments/assets/1497dfd5-143f-450b-88ff-9d18a5f16e4c" />
<img width="1912" height="914" alt="Image" src="https://github.com/user-attachments/assets/f8654e7b-fa3f-4fcf-afe5-361a4ac3d1c6" />

shiv3:
@CHENJIANMIN-git 
Thanks for using it and for the bug report.
Could you check if this fixes the issue?
https://github.com/shiv3/ocpp-cp-simulator/pull/104

CHENJIANMIN-git:
Yes, it is fixed. Thank you very much!

<img width="1912" height="914" alt="Image" src="https://github.com/user-attachments/assets/631a56a1-76fb-4f2a-a55a-198e6d35641e" />

CHENJIANMIN-git:
<img width="1912" height="914" alt="Image" src="https://github.com/user-attachments/assets/18013675-c0f1-4754-80c9-da59228588d9" />
Hi @shiv3 , it seems to reproduce after I pull the latest code. Please check it again, thanks!

```
commit 6c42d892348ed37cd2cd50c40cf242f6566b5ea4 (HEAD -> main, tag: v0.6.0, tag: v0.5.3, origin/main, origin/HEAD)
Merge: 6597cf4 d73e7a6
Author: shiv3 <tshibaza@gmail.com>
Date:   Fri Jul 3 00:40:53 2026 +0900
```

shiv3:
Thanks for the follow-up @CHENJIANMIN-git — reopening, you're right that it regressed.

The original #104 fix (retarget-on-import) is still in place, but the reproduction now comes from a **separate durability bug** in the browser storage layer (`SqlJsDatabase.flush()`, local/web-console mode only):

[...]

PR #112, opened 2026-06-30
refactor(scenario): centralize + sanitize scenario persistence (#101-class hardening)

shiv3:
Part of a comprehensive scenario-editor quality pass to stop the recent UI bug *classes* from recurring. This branch hardens **scenario persistence**.

## Problems rooted out
- **Autosave silent data loss in remote mode** — the autosave effect called `scenarioRepository.save` directly, which is a no-op in remote (daemon) mode, so edits vanished on reload. Same class as #101, but worse: no explicit user action.
- **Runtime state contaminating persisted data** — execution `progress`, highlight `style`/`className`, and live meter `currentValue` were written into `node.data` and then autosaved/exported, polluting the persisted source of truth.
- **Duplicated, mode-unaware save paths** — autosave/manual reimplemented the mode check instead of going through a single helper.

## Fix
- `serializeScenarioGraph` strips runtime/display overlays; the persistence helpers sanitize **at the boundary**, so no save/export path can persist them (the in-memory graph keeps overlays for display).
- `saveEditorScenario` = mode-aware **upsert (no prune)** for autosave + manual save; `persistEditorScenario` (prune) stays for import/template. Remote autosave runs through `createLatestWinsSaver` so a stale in-flight save can't clobber a newer one. Autosave fire timing is unchanged (same debounce/structural-key).
- All component save paths route through the helpers; an **ESLint denylist** forbids raw `scenarioRepository.save/load/delete` and `chargePointService.loadScenario/removeScenario` outside the helper, so the class can't silently return.

## Tests / gates
- New `scenarioSerialize.test.ts` (overlays never persist; real data/position preserved) + extended `scenarioPersistence.test.ts` (`saveEditorScenario` upsert-no-prune; latest-wins ordering). Full vitest **325 passed**. `tsc -b` no new errors. eslint clean.
- Codex-reviewed (blocker-free). Accepted limitation: the ESLint denylist catches direct calls, not aliased/destructured variants (guardrail, not a wall).

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -27,6 +27,50 @@ module.exports = {
   },
   overrides: [
     {
+      files: ["src/components/**/*.{ts,tsx}"],
+      rules: {
+        "no-restricted-syntax": [
+          "error",
+          {
+            selector:
+              "CallExpression[callee.type='MemberExpression'][callee.object.name='scenarioRepository'][callee.property.name='save']",
+            message:
+              "route scenario persistence through scenarioPersistence helpers (#101 class).",
+          },
+          {
+            selector:
+              "CallExpression[callee.type='MemberExpression'][callee.object.name='scenarioRepository'][callee.property.name='load']",
+            message:
+              "route scenario persistence through scenarioPersistence helpers (#101 class).",
+          },
+          {
+            selector:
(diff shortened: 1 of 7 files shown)


shiv3:
Folded into the single combined PR refactor/scenario-editor-quality (per request to consolidate into one PR). The commits are cherry-picked there.

Issue #127, opened 2026-07-08
Blank page in Docker image 0.6.1: Object.defineProperty called on non-object


juherr:
## Summary

The web UI bundled in Docker image `ghcr.io/shiv3/ocpp-cp-simulator:0.6.1` fails at startup with a blank page.

The same deployment works again after rolling back to `ghcr.io/shiv3/ocpp-cp-simulator:0.6.0`.

## Environment

- Image: `ghcr.io/shiv3/ocpp-cp-simulator:0.6.1`
- Previous working image: `ghcr.io/shiv3/ocpp-cp-simulator:0.6.0`
- Browser UI served from the container
- Browser: Chromium-based browser
- Deployment: Docker Compose behind an auth/reverse proxy

## Error

Browser console:

```text
index-W3H7gjIa.js:5227 Uncaught TypeError: Object.defineProperty called on non-object
    at Object.defineProperty (<anonymous>)
    at vmt (index-W3H7gjIa.js:5227:134789)
    at index-W3H7gjIa.js:5227:154489
    at Tt (index-W3H7gjIa.js:5227:154739)
    at m6 (index-W3H7gjIa.js:5227:2928)
    at hc (index-W3H7gjIa.js:5227:83167)
    at xn (index-W3H7gjIa.js:5227:85040)
    at F$ (index-W3H7gjIa.js:5227:101091)
    at index-W3H7gjIa.js:5227:106990
```

## Steps to reproduce

1. Run the Docker image:

```bash
docker run --rm -p 5172:5172 ghcr.io/shiv3/ocpp-cp-simulator:0.6.1
```

2. Open the web UI in a browser.
3. Observe that the page is blank.
4. Open DevTools console.

## Expected behavior

The web UI should load normally, as it does with Docker image `0.6.0`.

## Actual behavior

The UI stays blank and the browser console reports:

```text
Uncaught TypeError: Object.defineProperty called on non-object
```

## Additional findings

I verified that the failing bundle filename from the browser console, `index-W3H7gjIa.js`, is the bundle served by the `0.6.1` container, so this does not appear to be a browser cache issue.

Rolling back to `0.6.0` restores the UI immediately. The working `0.6.0` container serves a different bundle, `index-AVOu3U6P.js`.

The failure looks like a CommonJS interop / bundling initialization issue: `Object.defineProperty` appears to be called on a non-object, likely because a CommonJS `exports` object is not initialized when the bundled module executes.

[...]

PR #162, opened 2026-07-11
feat(ocpp): OCPP 1.2 and full 1.6-SOAP support, browser send-only SOAP (#91)

### LINKED ISSUE #91: Feature request: support legacy OCPP versions (1.5 / 1.2) for completeness
First, thanks for this simulator — with 0.4.0 covering 1.6J, 2.0.1 and 2.1, it already spans the whole modern OCPP stack, which is great.

**Suggestion:** add optional support (behind a flag/selector, like the existing version switch) for the older OCPP versions — **1.5** and **1.2**.

**Why it's worth considering:**

- **Completeness** — one tool covering the full OCPP lineage (1.2 → 2.1).
- **It fills a real gap.** As far as I can tell, there is *no* maintained simulator for the legacy SOAP-based versions. Anyone maintaining, debugging, or migrating a brownfield fleet still on 1.5 has nothing to test their CSMS against. That's a stronger argument than my own need, which is honestly more "nice to have completeness" than urgent.
- Directly relevant to the repo's stated "CSMS development" use case: validating backward-compatibility paths.
- **Ready-made companion CSMS:** [SteVe](https://github.com/steve-community/steve) already supports OCPP **1.2 / 1.5 / 1.6** (SOAP + JSON). So a legacy-capable CP simulator would pair directly with SteVe to give a complete legacy CP↔CSMS test loop out of the box — no need to build a CSMS to test against. (I already run SteVe alongside this simulator, so I'd happily help validate.)

**Caveats I'm aware of:** 1.2 and 1.5 are **SOAP**-based (JSON only arrived with 1.6), so this is a different transport than the current path — non-trivial work. If it's of interest, scoping to **1.5 first** (more widely deployed than 1.2) would already cover most of the value.

shiv3:
## Summary

Completes the remainder of #91 by adding **OCPP 1.2** and **OCPP 1.6 over SOAP (1.6S)**
alongside the existing OCPP 1.5 SOAP support, and extends the simulator so every version is
usable from the standalone browser (send-only for SOAP). All three SOAP versions share one
codec and, where applicable, the same domain handlers the WebSocket path already uses — so
this is largely a reuse/parameterization effort rather than new protocol machinery.

`OCPP-1.5` was already shipped; the new work here is 1.2 (narrower surface) and full 1.6-SOAP
(the complete 1.6 CS→CP command set with real domain effects).

Closes #91.

## What's included

**OCPP 1.2 (SOAP)** — `urn://Ocpp/{Cs,Cp}/2010/08/`
- Narrower surface per the 1.2 spec: 9 CP→CS + 9 CS→CP operations (no DataTransfer,
  GetConfiguration, Local Auth List or Reservation).
- 4-value ChargePointStatus (no `Reserved` → mapped to `Unavailable`), 8-value error-code set
  (unmapped 1.6 codes collapse to `Mode3Error`), flat `{timestamp,value}` MeterValues,
  StartTransaction without `reservationId`.

**OCPP 1.6 (SOAP) — full coverage** — `urn://Ocpp/{Cs,Cp}/2015/10/`
- The complete 1.6 message set over SOAP, with 1.6J-identical enums (no collapsing) and the
  1.6 `meterValue`/`sampledValue` child-element shape.
- CS→CP commands (TriggerMessage, RemoteStart/Stop, ChangeAvailability/Configuration,
  GetConfiguration, UnlockConnector, Reservation, Local Auth List, Smart Charging,
  GetDiagnostics, UpdateFirmware, DataTransfer, …) are answered with **real domain effects**
  by dispatching through the same v16 handler registry the WebSocket transport uses.

**Browser standalone (local mode)**
- All six versions are now selectable. SOAP versions run **send-only** in a browser tab
  (CP→CSMS calls over `fetch`; CSMS-initiated commands need the CLI/daemon that hosts the
  callback endpoint). The config modal explains the send-only/CORS constraint and only
  requires the callback URL in daemon/remote mode.
- The `xmlbuilder2`/`fast-xml-parser` browser-stub aliases (added for #127) are removed
  entirely — the codec is browser-safe now that it builds XML with a small hand-rolled,
  escaped string builder; the #127 bundle-scan regression guard stays in place.

**State-machine work**
- Per-version ChargePointStatus/errorCode projection is centralized in one domain module
  (the state machine keeps the 9-value 1.6 vocabulary; SOAP wire profiles consume the
  projection).
- The BootNotification gate, previously duplicated across the WebSocket and SOAP handlers,
  is unified into a shared `BootGate` (transport-specific queue-vs-drop behavior preserved).
- Scenario `RemoteStart`/`RemoteStop`/`Reservation` trigger nodes now **fail fast** with a
  clear message when the charge point's transport can never receive that CSMS call (send-only
  SOAP, 1.5's Reset-only registry, or an operation outside the dialect), instead of parking
  the scenario until timeout.

## Design

- A `SoapDialect` (version + CS/CP namespaces + operation surface) is the single point of
  variation between SOAP versions; the SOAP 1.2 + WS-Addressing envelope framing is shared.
  `OCPP-1.5` output is byte-identical to before.
- Extracting `buildV16CallHandlerRegistry()` lets the SOAP CS→CP server reuse the existing
  1.6 handlers; SOAP string payloads are coerced to typed request shapes via the vendored
  v16 JSON schemas before dispatch.

## Wire correctness

Namespaces, operation lists, enum sets and XSD element sequences were taken from the official
OCA 1.2 / 1.5 / 1.6 WSDLs (field order matters for JAXB-based servers). Notable verified
points: 1.6 `SampledValue` uses child elements (1.5 uses attributes); 1.2
`statusNotificationRequest` is `connectorId/status/errorCode` only; DataTransfer exists in
both directions in 1.5/1.6 (same wrapper, namespace-disambiguated); the 1.2 response enum
sets (e.g. `Scheduled` is valid, `RebootRequired` is not) drive the 1.6→1.2 response mapping.

## Testing

- `bun run lint`, `bun run format:check`, `vitest` (803 pass), `bun test bun.test` (129 pass)
  all green; golden snapshots byte-identical; `bun run build` succeeds without the removed
  stub aliases.
- New tests cover: per-dialect envelope round-trips, 1.2 narrow-shape/enum-collapse client
  output, 1.6S passthrough + child-element MeterValues, CS→CP registry dispatch and payload
  coercion, the 1.2 response-enum transform, `BootGate`, the status projection matrix, and
  the scenario capability fail-fast.

## Suggested follow-ups (not in this PR)

- Manual end-to-end verification of 1.2S / 1.6S against a live SOAP CSMS (the browser build's
  cross-origin behavior in particular is CORS-dependent).
- `xmlbuilder2` is now an unused dependency; dropping it is left out here to avoid an
  unrelated lockfile change.

https://claude.ai/code/session_01MctL5p6oN7CUA1KiCt4twC

## Summary by CodeRabbit

* **New Features**
  * Expanded SOAP support to OCPP **1.2** and **1.6S** across CLI, server routing, and UI configuration, with version-aware SOAP payload handling.
  * Improved SOAP **CSMS → charge point** dispatch for **OCPP 1.6** variants, including trigger/message workflows.
  * Enabled SOAP in browser/local mode for supported OCPP SOAP versions.
* **Bug Fixes**
  * Strengthened SOAP namespace/dialect validation and improved SOAP fault responses when configuration is missing.
  * Scenario waits now **fail fast** when CSMS-initiated SOAP capabilities aren’t available.
* **Documentation**
  * Updated README/server docs/CLI guidance for SOAP setup and endpoint versions (**1.2 / 1.5 / 1.6S**), including SteVe integration.


shiv3:
## Verified end-to-end against a live SteVe CSMS

Drove both new versions against a local SteVe (SOAP 1.2 + 1.6) over its
`CentralSystemService` endpoint. This surfaced — and fixed — a real interop bug
(`624110b`): SteVe emits the **fully-qualified** `wsa:Action`
(`urn://Ocpp/Cs/2010/08/CentralSystemService/BootNotificationResponse`) rather than the bare
`/BootNotificationResponse` the parser required, so every SOAP response was rejected and the
boot gate never opened. Pre-existing since the 1.5 client; the parser now matches on the
trailing operation segment (guarded by a new test).

**OCPP 1.2 (SOAP)** — full CP→CS lifecycle, every message `Accepted`, no Faults. SteVe DB:

```
transaction_pk=6  charge_box=CP12SOAP  evse=1  id_tag=TAGSOAP  start=1000  stop=4500  stopped=1
```

**OCPP 1.6 (SOAP)** — same lifecycle; SteVe DB recorded the transaction **and** the meter
values with full 1.6 metadata, confirming the `sampledValue` child-element shape parsed
correctly by SteVe's JAXB stack (attributes would have failed):

```
transaction_pk=7  charge_box=CP16SOAP  start=1000  stop=4500  stopped=1
meter_value:  4500  Energy.Active.Import.Register  Wh  Sample.Periodic
meter_value: 150000 Power.Active.Import            W   Sample.Periodic
```

**CS→CP (1.6S bidirectional)** — a SteVe-shaped `TriggerMessage(StatusNotification)` POSTed to
the daemon's callback endpoint was parsed (fully-qualified action + CP namespace), dispatched
through the shared v16 handler registry, answered `Accepted`, and the CP then sent the
triggered `StatusNotification` which SteVe accepted. The daemon's default scenario's
`RemoteStartTrigger` node correctly *waited* (rather than failing fast) on this server-hosted
CP, exercising the capability check live.

shiv3:
Thanks @coderabbitai — addressed in `6d0f1e9`:

- **Schema coercion** (`v16RegistryDispatch.ts`): `coerceSoapPayloadWithSchema` now recurses
  into array items via the schema's `items`, so numeric fields inside arrays
  (`chargingSchedulePeriod`'s `startPeriod`/`limit`, `localAuthorizationList` entries) are
  coerced instead of left as strings; object recursion is guarded to plain records, so a
  `null` or array value no longer throws or gets walked as an index-keyed record. Added tests
  for the array-of-objects, single-element-wrap, and null/array cases.
- **README**: dropped the internal "slice-1" label and stated the CSMS→CP endpoint answers
  Reset on every SOAP version plus the full 1.6 command set on 1.6S.
- **httpServer fallback fault message**: the "OCPP 1.5 SOAP" wording was already generalized to
  "OCPP SOAP" in `455cd20` (both the sibling check and the fallback), so no further change there.

Issue #179, opened 2026-07-11
feat(certification): add machine-readable expectations, verdicts, and reports to headless scenario runs


juherr:
## Current capabilities

v0.7.0 already provides a comprehensive headless scenario lifecycle through
the CLI and Socket.IO RPC API:

- `list_scenario_templates`
- `load_scenario_template`
- `run_scenario_template`
- `scenario_status`
- `stop_scenario`
- `step_scenario`
- `remove_scenario`
- scenario lifecycle events
- `logs.get`
- `state.reset`

This issue does not propose replacing or duplicating those APIs.

## Problem

The existing APIs report execution state, but they do not provide a
machine-verifiable certification result.

In particular:

- `scenario_status.state === "completed"` means graph execution completed,
  not that OCPP expectations were satisfied;
- a waiting execution exposes `currentNodeId`, but not a normalized description
  of the CSMS action or condition it expects;
- OCPP logs are available, but are not correlated into a scenario-run transcript;
- there is no structured PASS/FAIL verdict or JSON/JUnit campaign report;
- cleanup is either scenario-specific or global through `state.reset`.

As a result, an autonomous runner can launch the scenarios, but it must still
understand scenario JSON internals, interpret logs, determine the verdict, and
build reports itself.

## Proposal

Extend the existing scenario lifecycle with three complementary capabilities.

### 1. Structured expectations

When a scenario is waiting, expose the expected condition:

```json
{
  "scenarioId": "cert16-tc010-remote-start",
  "state": "waiting",
  "currentNodeId": "wait-for-remote-start",
  "expectation": {
    "type": "ocpp_call",
    "direction": "CSMS_TO_CP",
    "action": "RemoteStartTransaction",
    "constraints": {
      "connectorId": 1
    },
    "timeoutMs": 30000
  }
}
```

This could extend `scenario_status` without introducing a parallel lifecycle API.

### 2. Assertions and verdicts

Allow scenario definitions to declare assertions for:

- OCPP action and partial payload matching;
- request/response correlation;
- message ordering;
- state transitions;
- timeouts;
- absence of unexpected messages.

Expose a final verdict separately from execution state:

```json
{
  "executionState": "completed",
  "verdict": "PASS",
  "assertions": [
    {
      "id": "boot-accepted",
      "status": "passed"
    }
  ]
}
```

Suggested verdicts:

- `PASS`
- `FAIL`
- `BLOCKED`
- `SKIPPED`

### 3. Structured reports

[...]

Issue #111, opened 2026-06-30
Idea: turn the simulator into a turnkey OCPP test harness (WireMock-style) — stable API + Testcontainers


juherr:
Following up on #110. The daemon already exposes a **documented Socket.IO control API** ([docs/server.md](https://github.com/shiv3/ocpp-cp-simulator/blob/main/docs/server.md)) that goes well beyond just sending messages — it already covers the full lifecycle a test harness needs:

- **CP CRUD** via daemon methods: `cp.create` (`cpId`, `wsUrl`, `connectors`, `vendor`, `model`, `ocppVersion`, `basicAuth`, `bootNotification`, `autoConnect`), `cp.update`, `cp.delete`, `cp.list`.
- **Per-CP configuration & control**: `connect`/`disconnect`, connector management (`remove_connector`, `set_ev_settings`, `set_auto_meter_config`, `set_mode`, `set_soc`, …).
- **Full scenario lifecycle**: `load_scenario` / `run_scenario` / `step_scenario` / `scenario_status` / `stop_scenario`.
- **Assertions**: `event` push + rooms (`events.subscribe`) with typed (zod) contracts shared with the UI/CLI.

That's a great foundation, so this is more about **packaging and positioning** than net-new plumbing.

**The idea:** lean into the "scriptable OCPP stub" framing and make this the **WireMock of OCPP** — the turnkey, language-agnostic charge-point simulator you drop straight into a CSMS test suite. What makes WireMock turnkey isn't only its API; it's the ergonomics around it: a **Testcontainers** module, a JUnit extension, thin language clients, and declarative scenario files. The same recipe fits here perfectly.

**Concrete goal (ties into #110):** let any CSMS project *replace* its hand-coded OCTT certification ITs — e.g. SteVe's [`Ocpp16JsonCsmsCertificationIT`](https://github.com/steve-community/steve/tree/master/src/test/java/de/rwth/idsg/steve/certification/ocpp16) — with a combo of:

1. the **built-in official certification scenarios** (#110), and
2. a thin **client + Testcontainers** layer.

So in its own CI a CSMS project would just: spin up the simulator container, `cp.create` a charge point pointed at the CSMS-under-test, `run_scenario` `TC_xxx`, and assert pass/fail over the `event` stream — no per-project reimplementation of the CP side.

[...]

Issue #110, opened 2026-06-30
Feature request: ship the official OCPP 1.6 certification test cases as built-in scenarios


juherr:
The Scenario Editor is a superb addition — and great that it's right there in the web console. It opens up something that could be a killer feature:

**Suggestion:** pre-load the **official OCPP 1.6 certification test cases** as ready-to-run, built-in scenarios. Since the simulator *is* a charge point, it's the ideal tool to drive a CSMS through the CP side of each test case — live, straight from the UI.

**Why this would be valuable:**

- It gives CSMS developers a one-click way to exercise their server against the standard, well-known scenarios — the fastest possible feedback loop while building or debugging a CSMS.
- It's perfect flagship content for the Scenario Editor: users can run them as-is, then clone and tweak.
- The suite spans the Core, Local Auth List, Reservation, Remote Trigger, Smart Charging and Firmware Management profiles, so it exercises a broad slice of the protocol out of the box.

**References:**

- Official test cases (OCA): https://openchargealliance.org/certificationocpp/certification-ocpp-1-6/
- A coded reference — SteVe implements them as JUnit integration tests across Core / LocalAuth / Reservation / RemoteTrigger / SmartCharging / Firmware, e.g. `TC_001` Cold Boot, `TC_003`/`TC_004` regular charging session (plug-in / identification first), `TC_010`–`TC_012` remote start/stop, `TC_013`/`TC_014` hard/soft reset, `TC_017`/`TC_018` unlock connector, `TC_061` clear auth cache, etc.: https://github.com/steve-community/steve/tree/master/src/test/java/de/rwth/idsg/steve/certification/ocpp16

**License note:** SteVe is GPL-3.0 and this project is Apache-2.0, so its code is best used only as a reference for the expected message flows — not copied in. The test cases themselves come from the OCA spec, so reimplementing from the official definitions keeps the licensing clean.

**Scope:** even just the Core profile cases (cold boot, charging session, remote start/stop, reset, unlock, config retrieval) would already be hugely useful as a first pass; the other profiles can follow incrementally.

I'd be glad to help validate the scenarios against SteVe as the CSMS on my side — I already run it, so I can cross-check that the simulator's scenarios drive a real CSMS correctly.

goekay:
hey, maintainer of SteVe here. i am ready to support or help in whatever capacity.

juherr:
[...]

PR #210, opened 2026-07-14
docs(examples): Java + Testcontainers harness prototype (#111)

shiv3:
Addresses #111 (the "WireMock of OCPP" idea) with a concrete starting point.

## What this adds

`examples/testcontainers-java/` — a minimal prototype of driving the simulator as a **turnkey OCPP charge-point stub** from a JVM test suite. A CSMS project can spin up the simulator container, `cp.create` a charge point pointed at its CSMS-under-test, `run_scenario` a built-in certification scenario (#110), and assert the machine-readable verdict (#179) — without hand-coding the charge-point side per project.

`CertificationScenarioIT` drives the daemon entirely over the documented [Socket.IO control plane](../docs/server.md):

1. start the image with Testcontainers, wait on `GET /v1/healthz`;
2. connect Socket.IO, `cp.create`;
3. `events.subscribe`;
4. `load_scenario` (inline) + `run_scenario`;
5. poll `scenario_report`, assert `verdict == "PASS"`.

Self-contained: the CP's `wsUrl` points at a dead port, so it never connects to a CSMS and an `ocpp_absent` assertion for `Reset` is deterministically satisfied → PASS. Pointing `wsUrl` at a real CSMS-under-test and running a `TC_xxx` template is the same shape.

## Why it's trustworthy despite not running in CI

The Java is authored against `docs/server.md` and is **not** wired into this repo's (TypeScript) CI. To guard the exact flow it demonstrates, this PR also adds `src/cli/server/__tests__/harnessScenarioVerdict.bun.test.ts`, which drives the **identical RPC sequence against the real socket server** and asserts the PASS verdict — so the contract the example depends on is covered by CI.

## On #111's other asks

- The one named **code gap** — `securityProfile` / `authorizationKey` / SOAP callback config on `cp.create`/`cp.update` — is **already closed and wired** (schema in `src/protocol/methods.ts`, persisted + validated in `CPRegistry`). So a harness can already drive secured and SOAP certification scenarios.
- The **API-stability commitment** (is the Socket.IO control plane a versioned contract external harnesses can rely on?) is a maintainer decision, left open here.
- A full Testcontainers **module + thin client** is a natural follow-up; the issue author offered to prototype the Java side, and this is a foundation for that.

## Verification

- New bun test: green (drives cp.create → load_scenario → run_scenario → scenario_report → PASS over the real socket server).
- `bun test bun.test` — 217/0 (includes the new test).
- prettier / eslint clean. The Java prototype was not compiled here (no JVM in CI); it targets stable Testcontainers / socket.io-client APIs.

## Summary by CodeRabbit

- **New Features**
  - Added a Java + Testcontainers example harness for running end-to-end simulator certification scenarios.
  - Updated documentation with setup, prerequisites, execution steps, and configuration options for the harness.
  - Added an end-to-end control-plane flow (Socket.IO) that creates a charge point, runs an inline scenario, and validates the resulting verdict.

- **Tests**
  - Added Bun end-to-end coverage for the full certification scenario lifecycle and `PASS` verdict reporting.
  - Added a Java integration test using Testcontainers to exercise the same certification flow.

### DIFF
--- a/docs/server.md
+++ b/docs/server.md
@@ -269,6 +269,10 @@ await rpc({ method: "server.shutdown", params: {} });
 socket.disconnect();
 ```
 
+For a JVM/Testcontainers harness that drives this same control plane (start the
+container, `cp.create`, `run_scenario`, assert the machine-readable verdict), see
+[`examples/testcontainers-java/`](../examples/testcontainers-java/) (issue #111).
+
 ## Controlling a Running Daemon from the CLI
 
 A daemon does not have to be driven by a raw Socket.IO client or the web
--- a/examples/testcontainers-java/README.md
+++ b/examples/testcontainers-java/README.md
@@ -1,0 +1,63 @@
+# Java + Testcontainers harness (prototype)
+
+A minimal example of driving `ocpp-cp-simulator` as a **turnkey charge-point
+stub** from a JVM test suite — the "WireMock of OCPP" idea from
+[issue #111](https://github.com/shiv3/ocpp-cp-simulator/issues/111).
+
+A CSMS project can spin up the simulator container, create a charge point
+pointed at its CSMS-under-test, run a built-in certification scenario
+([#110](https://github.com/shiv3/ocpp-cp-simulator/issues/110)), and assert the
+machine-readable verdict ([#179](https://github.com/shiv3/ocpp-cp-simulator/issues/179))
+— without reimplementing the charge-point side per project.
+
+## What it does
+
+`CertificationScenarioIT` drives the daemon entirely over the documented
+[Socket.IO control plane](../../docs/server.md):
+
+1. start the simulator image with Testcontainers and wait on `GET /v1/healthz`;
+2. connect a Socket.IO client and `cp.create` a charge point;
+3. `events.subscribe` to that CP's event room;
+4. `load_scenario` (an inline scenario here) and `run_scenario`;
+5. poll `scenario_report` and assert `verdict == "PASS"`.
+
+The example is **self-contained**: the CP's `wsUrl` points at a dead port, so it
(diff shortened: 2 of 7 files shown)


juherr:
@shiv3 This is a very useful example, especially because it demonstrates how the simulator can be integrated into a CSMS test suite through Testcontainers and the documented control API.

I have a few suggestions that could make this approach even more maintainable and reusable over time.

### 1. Publish a JSON Schema for scenarios

The scenario format currently seems to be defined by the TypeScript model (`ScenarioDefinition`), but I couldn't find a versioned JSON Schema describing the external format.

Having a JSON Schema would bring several benefits:

- validate scenario files before loading or executing them;
- provide IDE autocompletion and documentation;
- clearly define the public contract independently of the TypeScript implementation;
- make the format easier to consume from other languages;
- provide a solid foundation for a future scenario DSL or fluent API (Java, Kotlin, Go, etc.).

It might also be worth adding a `schemaVersion` field so the format can evolve without breaking compatibility.

### 2. Execute the examples in CI

Would it be possible to include this example (and future examples) in the CI pipeline?

Examples tend to become outdated as APIs evolve. Running them automatically would ensure they remain valid over time and would effectively turn them into compatibility tests.

Ideally the CI could:

- build the example;
- start the simulator container;
- execute `CertificationScenarioIT`;
- verify the expected `PASS` verdict.

That would also help catch accidental breaking changes to the control API, scenario format or report structure.

### 3. Use the Maven Wrapper

For the Maven example, I'd also recommend committing the Maven Wrapper (`mvnw`, `mvnw.cmd` and `.mvn/wrapper/`).

That way both contributors and CI can simply run:

```bash
./mvnw verify
```

without requiring a specific Maven installation, making the example more reproducible.

Issue #188, opened 2026-07-12
Collaborate with OCPP DebugKit on a shared trace format and integrated analysis UI


juherr:
## Context

`ocpp-cp-simulator` already provides structured OCPP logs, scenario execution, automation capabilities, JSONL output, a browser interface, and CI-oriented tooling.

In parallel, [`ocpp-debugkit/toolkit`](https://github.com/ocpp-debugkit/toolkit), maintained by @sepehr-safari, provides complementary capabilities such as:

- OCPP trace parsing and normalization
- session timeline reconstruction
- known failure-pattern detection
- deterministic trace replay
- Markdown and HTML report generation
- reusable React inspection components

The two projects appear highly complementary:

- `ocpp-cp-simulator` generates and controls OCPP exchanges;
- OCPP DebugKit analyzes, explains, and visualizes recorded exchanges.

OCPP DebugKit already includes an example for processing simulator JSONL output, which suggests that a more formal collaboration could be valuable.

## Proposal

Would @sepehr-safari and @shiv3 be interested in collaborating on a shared trace format and a progressive integration of selected DebugKit capabilities into the simulator?

The proposal could be split into two objectives.

## Primary objective: define a common, versioned OCPP trace format

The first step could be to agree on a small, implementation-independent JSON or JSONL format capable of representing:

- schema version
- OCPP version
- transport, such as JSON/WebSocket or SOAP/HTTP
- charge point identity
- connector and EVSE identity when applicable
- message direction
- request, response, and error message types
- action name
- message identifier and correlation identifier
- timestamp
- raw payload
- optional transport and execution metadata
- optional simulator or analysis-specific extensions

For example:

```json
{
  "schemaVersion": "1.0",
  "timestamp": "2026-07-12T14:32:10.123Z",
  "ocppVersion": "1.6",
  "transport": "json",
  "chargePointId": "CP001",
  "direction": "cp-to-csms",
  "messageType": "CALL",
  "messageId": "abc-123",
  "action": "BootNotification",
  "payload": {
    "chargePointVendor": "Example",
    "chargePointModel": "Simulator"
  }
}
```

The exact structure should be discussed rather than assumed.

A shared format could be useful beyond these two projects, for example for:

- simulator interoperability
- CSMS debugging
- reproducible bug reports
- CI artifacts
- regression test fixtures
- offline analysis
- comparison of different CSMS implementations
- sharing traces between tools without coupling their internal models

[...]

PR #211, opened 2026-07-14
feat(trace): versioned OCPP trace format + log adapter (#188 PoC)

shiv3:
Addresses [#188](https://github.com/shiv3/ocpp-cp-simulator/issues/188) — the first step of its proof of concept: **document and version a shared OCPP trace format**, and adapt the simulator's own logs into it.

## What this adds

A small, **implementation-independent** trace format (`schemaVersion "1.0"`) for one OCPP message exchange, so simulator output can be consumed by external analysis tools (e.g. [OCPP DebugKit](https://github.com/ocpp-debugkit/toolkit)), shared as CI artifacts, or used for reproducible bug reports — **without coupling either side's internal models**.

- `src/trace/OcppTraceRecord.ts` — the versioned record type + `OCPP_TRACE_SCHEMA_VERSION`. Fields follow #188's proposal: `schemaVersion`, `timestamp`, `ocppVersion`, `transport`, `chargePointId`, `connectorId`, `direction`, `messageType`, `messageId`, `action`, `payload`, `error`, `meta`.
- `src/trace/logEntryToTrace.ts` — an adapter **from** the simulator's JSONL log lines (`--log-format json` / `logs.get` / the log-viewer download) **to** neutral trace records. `logLinesToTrace` drops non-wire lines and back-fills each CALLRESULT/CALLERROR `action` from the correlated CALL by `messageId`; `logLineToTraceRecord` handles a single line.
- `docs/trace-format.md` — field table, example, and a JSON Schema.
- Unit tests for the adapter.

## Scope (deliberately a PoC)

- **OCPP-J (JSON/WebSocket)** frames only in this iteration; SOAP is a documented follow-up (the format already carries a `transport` field).
- Producing trace files from a run (a `--trace-output` flag / an `analyze` subcommand) and any DebugKit integration are **out of scope** here, per the issue's staged plan.
- Direction is normalized to `cp-to-csms` / `csms-to-cp`; the format stays usable without this simulator or DebugKit.

## On the open questions in #188

Schema ownership (this repo vs. a dedicated repo vs. a published JSON Schema package), whether to depend on `@ocpp-debugkit/toolkit` directly or via an adapter, and the UI/analysis integration are **strategic/collaboration decisions** left to the maintainers and the DebugKit side — this PR just lands a concrete, versioned starting point to discuss against.

## Verification

- `vitest` — new adapter tests green (8).
- `bun test bun.test` — unaffected.
- prettier / eslint clean.

## Summary by CodeRabbit

* **New Features**
  * Introduced a standardized, versioned “OCPP Trace Format” contract for sharing trace records as JSON/JSONL.
  * Added a converter that transforms simulator WebSocket log lines into structured trace records, including transport, direction, message type, correlation, payload/action, and error details.
  * Automatically correlates CALL with subsequent CALLRESULT/CALLERROR, scoped correctly per charge point.
* **Documentation**
  * Added a new documentation spec with required/optional fields, examples, and validation rules via JSON Schema.
* **Tests**
  * Added test coverage for parsing, correlation, filtering of non-wire lines, and error mapping.

### DIFF
--- a/docs/trace-format.md
+++ b/docs/trace-format.md
@@ -1,0 +1,119 @@
+# OCPP Trace Format (v1.0)
+
+A small, **implementation-independent** JSON/JSONL format for one OCPP message
+exchange. It is intended as a shared contract between tools that _produce_ OCPP
+traffic (this simulator) and tools that _analyze_ it (e.g.
+[OCPP DebugKit](https://github.com/ocpp-debugkit/toolkit)), without coupling
+either side to the other's internal models — see
+[issue #188](https://github.com/shiv3/ocpp-cp-simulator/issues/188).
+
+This is a **proof of concept**: it defines and versions the record shape and
+ships an adapter from the simulator's own logs. Producing trace files from a run
+(a `--trace-output` flag / an `analyze` subcommand) and any DebugKit integration
+are deliberately later steps.
+
+## Status & scope
+
+- **Version `1.0`** (`schemaVersion`). Bump on any breaking change.
+- This iteration recognizes **OCPP-J (JSON/WebSocket)** frames. SOAP transport
+  is a documented follow-up (the format already carries a `transport` field).
+- The format is usable without this simulator or DebugKit; it is a plain JSON
+  object per exchange (JSONL for a stream).
+
+## Record shape
+
+| Field           | Type                                           | Notes                                                                              |
+| --------------- | ---------------------------------------------- | ---------------------------------------------------------------------------------- |
+| `schemaVersion` | string                                         | Trace schema version, e.g. `"1.0"`.                                                |
+| `timestamp`     | string (ISO-8601)                              | When the message was observed.                                                     |
+| `ocppVersion`   | string (optional)                              | OCPP protocol version, e.g. `"1.6"`, `"2.0.1"`.                                    |
+| `transport`     | `"json"` \| `"soap"`                           | Wire transport.                                                                    |
+| `chargePointId` | string (optional)                              | Charge-point identity.                                                             |
+| `connectorId`   | number (optional)                              | Connector id when connector-scoped and known.                                      |
+| `direction`     | `"cp-to-csms"` \| `"csms-to-cp"`               | Relative to the CP/CSMS pair.                                                      |
+| `messageType`   | `"CALL"` \| `"CALLRESULT"` \| `"CALLERROR"`    | OCPP-J frame kind.                                                                 |
+| `messageId`     | string (optional)                              | Correlates a CALL with its CALLRESULT/CALLERROR.                                   |
+| `action`        | string (optional)                              | e.g. `"BootNotification"`. On CALLRESULT/CALLERROR, back-filled by id correlation. |
+| `payload`       | any (optional)                                 | The OCPP message body.                                                             |
+| `error`         | `{ code?, description?, details? }` (optional) | Populated for CALLERROR only.                                                      |
+| `meta`          | object (optional)                              | Transport/execution metadata and analysis-specific extensions.                     |
+
+## Example
+
+```json
+{
+  "schemaVersion": "1.0",
+  "timestamp": "2026-07-14T02:00:00.000Z",
+  "ocppVersion": "1.6",
+  "transport": "json",
+  "chargePointId": "CP001",
+  "direction": "cp-to-csms",
+  "messageType": "CALL",
+  "messageId": "abc-1",
+  "action": "BootNotification",
+  "payload": { "chargePointVendor": "Example", "chargePointModel": "Simulator" }
+}
+```
+
+## Producing records
+
+The adapter in [`src/trace/`](../src/trace/) maps this simulator's JSONL log
+lines (`--log-format json`, the `logs.get` RPC, or the browser log-viewer
+download) into trace records:
+
+```ts
+import { logLinesToTrace } from "../src/trace/logEntryToTrace";
+
+const records = logLinesToTrace(jsonlLogLines, { ocppVersion: "1.6" });
+```
+
+`logLinesToTrace` drops non-wire log lines and back-fills each
+CALLRESULT/CALLERROR `action` from the CALL that established its `messageId`.
+`logLineToTraceRecord` converts a single line (returning `null` for non-wire
+lines). The type lives in `src/trace/OcppTraceRecord.ts` and
+`OCPP_TRACE_SCHEMA_VERSION` is the current version string.
+
+## JSON Schema
+
+```json
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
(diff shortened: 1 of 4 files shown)


juherr:
I’d be very interested in getting feedback from maintainers of other OCPP tools:

* @geokay from SteVe
* @sepehr-safari from OCPP DebugKit
* @pbourseau from ocpp-toolkit

This PR introduces a small, versioned, implementation-independent JSON/JSONL format for representing OCPP message traces.

The goal is to make traces portable between simulators, CSMS implementations, debugging tools, CI pipelines, and reproducible bug reports, rather than creating another format specific to this simulator.

From the perspective of your respective projects:

* Does the proposed format cover the information you would need to import, export, or analyze OCPP traces?
* Are there fields or design choices that would make adoption difficult?
* Would you consider implementing support for this format as either a producer or a consumer?
* Would it make sense to maintain the schema outside this repository, for example in a dedicated repository or package shared by the different projects?

The current version is intentionally a proof of concept focused on OCPP-J. Feedback at this stage would be particularly useful before the format becomes more established.

sepehr-safari:
Thanks for the ping @juherr. And @shiv3, nice work getting something concrete landed this fast, it's a lot easier to react to running code than to a proposal.

The framing in #188 matches how I think about it: simulators generate exchanges, DebugKit analyzes them. Keeping the coupling at a file format, rather than at each other's internal models or UI, is the right call.

Going through your four questions.

## 1. Does the format cover what we'd need?

Mostly yes. DebugKit's internal event model after normalization is already close to your record: id, message id, timestamp, direction, message type, action, payload, error fields. So we don't really disagree about what an analyzed message is.

`chargePointId`, `connectorId`, `transport` and `ocppVersion` are things we either infer or keep as trace-level metadata today. Having them per-record is better, especially for multi-CP traces.

One thing worth calling out as right: collapsing "message identifier" and "correlation identifier" into a single `messageId`. In OCPP-J they're the same thing, the CallResult echoes the Call's id. Two fields would just give people a way to disagree with themselves.

The gap for me is malformed messages, which is my main comment below.

## 2. Anything that would make adoption difficult?

One real thing, one minor.

### The record decomposes the OCPP-J frame, and I'd keep the raw array too

DebugKit keeps the frame verbatim (`"message": [2, "id", "BootNotification", {...}]`) and only decomposes in memory. That's on purpose, and the reason is specific to debugging: the messages you most want to move between tools are usually the broken ones. A Call with 3 elements instead of 4, a truncated frame, an unknown action, a payload that doesn't match the 1.6 schema. That's what people file reproducible bug reports about. Our parser treats those as untrusted input and skips-and-flags them rather than dying ([ADR-0007](https://github.com/ocpp-debugkit/toolkit/blob/main/docs/adr/0007-malformed-trace-handling.md)), but that only works because the format can carry them at all.

A typed `{messageType, messageId, action, payload}` record assumes the frame was well-formed. It also can't round-trip cleanly:

* CallError is `[4, id, errorCode, errorDescription, details]`, so `error` needs a defined shape or you lose fields.
* Key order and exact bytes are gone, which matters if anyone wants to hash, dedupe, or reproduce a frame exactly.

What I'd suggest: keep the decomposed fields, they're the useful part and they're what consumers actually read. Add an optional `raw` holding the original array. Producers that have the bytes emit it, ones that don't, don't. Consumers that care use it, ones that don't, ignore it. You keep the ergonomics and it stops being lossy for the debugging case.

### Minor: the back-filled `action` on CALLRESULT/CALLERROR

This PR back-fills `action` from the correlated CALL at serialization time. We do the same correlation, but at analysis time. Baking it into the record means a record can contradict the correlation it came from: if a producer gets messageId correlation wrong, the trace now asserts something false rather than just being incomplete. It is convenient though, so if you keep it I'd mark it derived and optional, and spell out that it MUST equal the action of the CALL with the same `messageId` when that CALL is present in the trace.

Not worth arguing about: `cp-to-csms` / `csms-to-cp` is fine. We use `CS_TO_CSMS` / `CSMS_TO_CS` internally, happy to map.

## 3. Producer or consumer?

Both.

Consumer is the easy one, `parseTrace()` already takes JSONL so it's an input adapter.

Producer is the more interesting one. DebugKit Studio (the native side of the project) proxies WebSocket traffic between a charge point and a CSMS and records live sessions. That's a natural emitter for this, and it's why I care about `raw`: it currently writes the exact bytes off the wire so that re-parsing the recording offline gives you an identical event.

I'd want to pin a `schemaVersion` and not chase it, so question 4 matters more to me than the field list does.

## 4. Should the schema live outside this repo?

Yes, and I'd push it further: a shared format needs a conformance suite or it forks quietly.

I have some direct experience here so I'd rather offer it than just assert it.

DebugKit is two independent implementations of the same analysis behavior, a TypeScript library and a native Zig app. They share no code. They only agree because there's a machine-checked contract between them: vendored trace fixtures, each pinned to the exact output the reference produces, re-run in CI on every change, release blocked if they disagree. That's the only reason "one format, two implementations" is something I can check rather than something I hope is true.

The usual failure mode is that a schema gets agreed, four projects implement it slightly differently, nobody notices for a year, and by then the format is fiction. A JSON Schema catches shape errors. It won't catch "your producer and my consumer disagree about what `direction` means on a CallError".

So concretely:

* Schema in a neutral repo, not this one and not DebugKit's, so nobody's release cadence or roadmap governs it.
* Ship fixtures plus expected parsed output next to the JSON Schema, so producers and consumers can self-check in CI.
* Versioned and immutable within a version. A change that alters what a conformant implementation outputs is a new version, not an edit. We freeze ours that way: 2.0.1 will be a new contract version rather than a change to the current one, so a pinned consumer can't break silently.

I'm happy to do the conformance suite and the harness, and to seed the corpus with DebugKit's 15 scenario traces (normal session, failed auth, connector fault, offline during session, timeout on transaction start, and so on). No preference on where it lives, and I'd honestly rather it not sit under `ocpp-debugkit`, since neutral ground is the whole point. If SteVe and ocpp-toolkit want in as well, that's several independent implementations checking each other against the same fixtures, which would be a good place to be.

Two caveats on what I can actually bring. Our contract is OCPP 1.6J only right now, and it has no concept of `transport` (no SOAP) or EVSE identity. Your proposal is broader than what we've shipped, so take the fixtures and the mechanism, not our field list.

If it's useful I can send a follow-up PR with the `raw` field and a defined CallError shape to make it concrete, or sketch out what the conformance repo would look like. Whichever is more helpful.

shiv3:
Replied on #188 to keep the discussion in one place — short version: `raw` yes, landed in #216 (merged) as a verbatim string (rather than an array, for truncated frames and byte-exact hashing) together with the derived/optional `action` constraint and explicit versioning rules, so the fixture conversion isn't blocked on it. `--trace-output` shipped in #217 as the first CLI piece. One open point flagged there: fully-unparseable frames still can't be records (`messageType` is required) — proposing we settle that in the spec repo.

PR #216, opened 2026-07-16
feat(trace): trace format v1.1 — raw frames, derived action, versioning rules (#188)

shiv3:
Follow-up to #211, addressing the format review feedback from the DebugKit side on [#188](https://github.com/shiv3/ocpp-cp-simulator/issues/188) / [#211 (review)](https://github.com/shiv3/ocpp-cp-simulator/pull/211#issuecomment-4969245960): trace format **v1.1**.

## What changes

Three additions to the shared trace contract, all additive (v1.0 records remain valid v1.1 records):

- **`raw` (optional string)** — the verbatim frame text exactly as sent/received on the wire. Producers that have the original bytes SHOULD emit it; it is the only lossless representation (byte-exact hashing/dedup, malformed frames). It is a *string*, not a parsed array: a truncated frame isn't valid JSON, so an array-typed `raw` couldn't carry exactly the malformed frames it exists for, and byte-exact hashing only works on the original text. Well-formed frames still yield the OCPP-J array via `JSON.parse(raw)`. The log adapter now fills it (the frame text after the `Sent: ` / `Received: ` prefix, untouched).
- **`action` semantics spelled out** — on CALLRESULT/CALLERROR it is *derived* by message-id correlation and *optional*; when the CALL with the same `messageId` is present in the trace it MUST equal that CALL's action. A producer that correlates wrongly now violates the spec rather than silently asserting a false fact.
- **Versioning rules made explicit** — additive optional fields bump the minor version; consumers MUST ignore unknown fields; changing the meaning of a field (or removing one) is a new major version; a published version is immutable (a change that alters conformant output is a new version, not an edit); producer extensions go in `meta`, never in undeclared top-level fields. Consequently the JSON Schema's `additionalProperties` flips to `true` so a pinned consumer can validate records from a later minor without rejecting them.

`OCPP_TRACE_SCHEMA_VERSION` is now `"1.1"`; `docs/trace-format.md` gains the Versioning section, updated field table/example/schema (`$id` → `#v1.1`), and a changelog.

## Why now

The fixture conversion planned on #188 (DebugKit's 15 scenario traces into the shared format) is blocked on the `raw` decision — landing v1.1 here lets the future `open-ocpp-trace/specification` repo seed from v1.1 as-is instead of starting on a known-stale v1.0.

## Verification

- `vitest` — 1323/1323 green (4 new v1.1 tests, incl. byte-exact preservation of non-canonical spacing/key order and `JSON.parse(raw)` round-trip).
- `bun test bun.test` — 221/221 green.
- eslint clean.

## Summary by CodeRabbit

* **New Features**
  * Trace records can now include the original, verbatim OCPP-J frame text via an optional `raw` field (when available).
  * The trace format is updated to version 1.1, including clarified `action` semantics for `CALLRESULT`/`CALLERROR`.
* **Documentation**
  * Expanded trace format documentation with updated versioning, forward-compatibility guidance, record-shape details, and schema updates.
* **Tests**
  * Added assertions to verify `raw` preservation and correct schema version handling across trace outputs.

### DIFF
--- a/docs/trace-format.md
+++ b/docs/trace-format.md
@@ -31,25 +31,29 @@ are deliberately later steps.
   conformant implementation outputs is a new version, not an edit.
 - Producer extensions MUST go in `meta`, never in undeclared top-level
   fields.
+- Known v1.1 limitation: a frame that does not parse as an OCPP-J array
+  cannot be represented as a record (`messageType` is required). How the shared
+  format should carry fully-unparseable frames is an open question for the
+  specification repo.
 
 ## Record shape
 
-| Field           | Type                                           | Notes                                                                                                                                         |
-| --------------- | ---------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------- |
-| `schemaVersion` | string                                         | Trace schema version, e.g. `"1.0"`.                                                                                                           |
-| `timestamp`     | string (ISO-8601)                              | When the message was observed.                                                                                                                |
-| `ocppVersion`   | string (optional)                              | OCPP protocol version, e.g. `"1.6"`, `"2.0.1"`.                                                                                               |
-| `transport`     | `"json"` \| `"soap"`                           | Wire transport.                                                                                                                               |
-| `chargePointId` | string (optional)                              | Charge-point identity.                                                                                                                        |
-| `connectorId`   | number (optional)                              | Connector id when connector-scoped and known.                                                                                                 |
-| `direction`     | `"cp-to-csms"` \| `"csms-to-cp"`               | Relative to the CP/CSMS pair.                                                                                                                 |
-| `messageType`   | `"CALL"` \| `"CALLRESULT"` \| `"CALLERROR"`    | OCPP-J frame kind.                                                                                                                            |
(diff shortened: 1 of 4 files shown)


shiv3:
Addressed in 0373ce3:

- Nested `error` schema now has `additionalProperties: true` (forward-compat note extended to cover it).
- Stale `"1.0"` examples updated to `"1.1"`.
- On the malformed-frame contract mismatch: took the narrow-the-docs route. `raw` is now documented as preserving frames whose *shape or payload* violates the OCPP schema (extra/missing elements, nonconformant payloads — these still parse as OCPP-J arrays and become records). A frame that does not parse as an OCPP-J array at all cannot be a record in v1.1 because `messageType` is required — that limitation is now stated explicitly in the Versioning section and the `raw` JSDoc, and flagged as an open question for the shared specification (whether `messageType` becomes optional when `raw` is present, a dedicated malformed kind, etc.). Implementing an opaque-record path here would pre-empt that spec decision.

shiv3:
@coderabbitai review

Issue #121, opened 2026-07-07
GitHub Pages deployment fails: deployment_queued until timeout


juherr:
Hi,

The latest GitHub Pages deployment for commit `d25916fee3cb3360b15fd66c150bca82e17d1b99` is failing.

Commit: https://github.com/shiv3/ocpp-cp-simulator/commit/d25916fee3cb3360b15fd66c150bca82e17d1b99  
Failed run: https://github.com/shiv3/ocpp-cp-simulator/actions/runs/28602703337  
Published app: https://shiv3.github.io/ocpp-cp-simulator/

From the logs, the generated artifact itself looks valid:

- `index.html`, `404.html`, `.nojekyll` and the bundled assets are included.
- The artifact is uploaded successfully.
- The artifact size is around 1 MB.

The failure happens later in `actions/deploy-pages@v5`.

The deployment is created successfully:

```text
Created deployment for d25916fee3cb3360b15fd66c150bca82e17d1b99
```

But then the deployment remains stuck in:

```text
Current status: deployment_queued
```

until the workflow reaches the 10-minute timeout:

```text
Timeout reached, aborting!
Canceling Pages deployment...
Canceled deployment with ID d25916fee3cb3360b15fd66c150bca82e17d1b99
```

So this does not look like a Vite build issue or a missing generated asset. It looks more like the GitHub Pages deployment is stuck/cancelled after the artifact upload.

One thing that may be worth checking: the repository seems to use `peaceiris/actions-gh-pages@v3` to push the built `dist` output to the `gh-pages` branch, and then GitHub Pages automatically runs `pages-build-deployment` on that branch. It may be worth verifying whether the Pages configuration and the deployment workflow are using a single consistent deployment strategy.

Possible fixes to try:

- Re-run the failed `pages-build-deployment` workflow.
- Push an empty commit to `gh-pages` to trigger a new Pages deployment.
- Check whether GitHub Pages is configured to deploy from the `gh-pages` branch while another workflow also tries to deploy Pages artifacts.
- Consider switching to the official GitHub Pages flow using `actions/upload-pages-artifact` + `actions/deploy-pages`, or keeping the `gh-pages` branch deployment, but avoiding mixed deployment mechanisms.

Thanks for maintaining this project.

shiv3:
Fixed in #124 (merged to `main` via #125).

[...]

PR #227, opened 2026-07-21
fix(release): attach desktop binaries to the published release, not a new draft

shiv3:
## Problem

Every release ships the **previous** version's desktop binaries.

| release | attached binaries |
|---|---|
| v0.7.2 | ❌ `0.7.1` × 8 |
| v0.7.1 | ❌ `0.7.0` × 8 |
| v0.7.0 | ❌ none |
| leftover draft (unpublished) | `0.7.2` × 8 |

## Root cause

Two mechanisms were both managing "the draft release".

The intended flow: release-drafter keeps a rolling draft on `main`; publishing it creates the `vX.Y.Z` tag; that tag triggers this workflow. **So by the time the desktop build runs, the release is already published.**

But tauri-action had `releaseDraft: true`. Rather than attaching to that published release, it created a *new* draft and parked the binaries there. Release Drafter then adopted the leftover draft on the next merge to `main` — it updates the newest existing draft rather than creating its own — renamed it to the next version, and the operator published it, binaries and all, one version stale.

### Evidence, not inference

- The assets on the published v0.7.2 were uploaded **08:19–08:23**, inside the **v0.7.1** build window (08:14:56–08:24:07). The v0.7.2 build (09:13–09:22) uploaded nothing to it.
- That release carries release-drafter's `v0.7.2` name (its template is `v$RESOLVED_VERSION`) while holding tauri-action's output — whose own naming is `OCPP CP Simulator $tag`. So the draft really was created by one mechanism and renamed by the other.
- The still-unpublished draft holding the correct `0.7.2` binaries is named `OCPP CP Simulator v0.7.2` — tauri-action's format, not yet renamed because no merge to `main` has happened since.
- v0.7.0 has no assets because it was the first turn of the cycle, with no leftover draft to inherit.

Same class of bug as #121, where two mechanisms both drove Pages deploys.

> Careful reading the API: a release's `created_at` is the **tagged commit's date**, not when the release object was created. The timeline looks contradictory until you account for that.

## Fix

- `releaseDraft: false` — tauri-action uploads into the release the tag already points at, leaving nothing for the drafter to adopt.
- Drop `releaseBody` — release-drafter owns the notes (its template already emits `## What's Changed`, which the `update-release-notes` job keys off to inject the commit list). A body passed here would clobber those categorized notes with a static blurb.
- The download table that lived in `releaseBody` moves into the drafter template, so it lands on every release regardless of which mechanism created it.
- Refresh the now-stale comment in `update-release-notes` about drafts being invisible to `getReleaseByTag()`.

## Verification

**Static only** — both files parse as YAML and are prettier-clean. I could not exercise the behaviour without cutting a real release, so I'm not claiming it's proven.

The next release is the test. Watch for:
1. binaries land on the published release itself, with matching version numbers
2. no stray draft is left behind afterwards
3. the release notes keep release-drafter's categorized `## What's Changed` plus the injected commit list

## Not included

The already-published v0.7.2 still carries 0.7.1's binaries. The correct `0.7.2` assets exist in the leftover draft, so fixing it needs no rebuild — but swapping assets on a public release is outward-facing and hard to undo, so I've left it for a deliberate decision rather than bundling it here.

## Summary by CodeRabbit

- **Documentation**
  - Release notes now link directly to the web version.
  - Updated the release template to add a “Desktop App Downloads” section and a “Web Version” subsection, with clearer artifact guidance for macOS (Intel/Apple Silicon), Windows, and Linux.
- **Chores**
  - Improved the release workflow to resolve/create releases before building, ensuring download links and notes are included reliably.
  - Made release-note updates more resilient by updating or adding the “What’s Changed” section when it’s missing.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -69,7 +69,7 @@ template: |
 
   - **macOS (Apple Silicon)**: `OCPP.CP.Simulator_*_aarch64.dmg`
   - **macOS (Intel)**: `OCPP.CP.Simulator_*_x64.dmg`
-  - **Windows**: `OCPP.CP.Simulator_*_x64-setup.exe` or `.msi`
+  - **Windows**: `OCPP.CP.Simulator_*_x64-setup.exe` or `OCPP.CP.Simulator_*_x64_en-US.msi`
   - **Linux (Debian/Ubuntu)**: `OCPP.CP.Simulator_*_amd64.deb`
   - **Linux (Fedora/RHEL)**: `OCPP.CP.Simulator-*.x86_64.rpm`
 
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -61,5 +61,21 @@ template: |
 
   **Full Changelog**: https://github.com/$OWNER/$REPO/compare/$PREVIOUS_TAG...v$RESOLVED_VERSION
 
+  ### Desktop App Downloads
+
+  Attached by release.yml a few minutes after this release is published — the
+  tag created on publish is what starts the desktop builds, so the assets show
+  up shortly after, not instantly.
+
+  - **macOS (Apple Silicon)**: `OCPP.CP.Simulator_*_aarch64.dmg`
+  - **macOS (Intel)**: `OCPP.CP.Simulator_*_x64.dmg`
+  - **Windows**: `OCPP.CP.Simulator_*_x64-setup.exe` or `.msi`
+  - **Linux (Debian/Ubuntu)**: `OCPP.CP.Simulator_*_amd64.deb`
+  - **Linux (Fedora/RHEL)**: `OCPP.CP.Simulator-*.x86_64.rpm`
+
+  ### Web Version
+
+  https://shiv3.github.io/ocpp-cp-simulator/
+
   ---
   Container image: `ghcr.io/$OWNER/$REPO:$RESOLVED_VERSION` (published when this release is tagged).
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,63 @@ on:
   workflow_dispatch:
 
 jobs:
+  # Resolve which release the binaries belong to, ONCE, before the matrix fans
+  # out. tauri-action is then handed a `releaseId` and acts as a pure asset
+  # uploader — see the comment on its step below for why it must not be allowed
+  # to resolve the release itself.
+  resolve-release:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    outputs:
+      release_id: ${{ steps.find.outputs.release_id }}
+    steps:
+      - name: Find (or create) the release for this tag
+        id: find
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            // workflow_dispatch runs off a branch, not a tag. Leave the id
+            // empty: tauri-action then has neither releaseId nor tagName and
+            // logs "skipping all uploads", so a manual run still smoke-builds
+            // the desktop app without inventing a release for a branch ref.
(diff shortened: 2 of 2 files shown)


shiv3:
Both findings were valid. Addressed in 130d830.

**🟠 Major — release must exist before `tauri-action` runs**

Confirmed, and it's a regression my own change introduced: dropping `releaseBody` removed the `## What's Changed` heading that the direct-tag-push path was relying on. Publishing a drafter release is the normal flow and unaffected, but a directly-pushed tag makes `tauri-action` create the release with an empty body, and `update-release-notes` only injected commits when that heading was already present — so the release would have ended up with no notes.

I didn't take the "fail the job when the release is missing" option, though. Tagging directly is unusual but it shouldn't block the desktop builds from producing binaries. Adding the heading when it's absent fixes the same hole, keeps both paths working, and is a smaller change.

Verified by running the actual snippet — extracted from `release.yml` at test time so it can't drift from the implementation — over four cases:

| input body | result |
|---|---|
| drafter body (heading present) | notes preserved, still one heading, commits slotted under it |
| empty (direct tag push) | heading + commit list added, no leading blank lines |
| prose without the heading | section appended cleanly |
| heading present, zero commits | body untouched |

**🟡 Minor — Windows MSI pattern**

Correct. The real asset name is `OCPP.CP.Simulator_0.7.2_x64_en-US.msi`, so a bare `.msi` was imprecise next to the explicit `.exe` pattern. Now spelled out. (That line was carried over verbatim from the old `releaseBody`, not new here — but the actual filenames were right there in the v0.7.2 release, so no reason to leave it vague.)

---

One thing worth flagging for whoever merges this: the fix stops *new* stray drafts, but it doesn't clean up the existing one. Draft `357226553` still holds the `0.7.2` binaries, and Release Drafter will adopt it on the next merge to `main` and rename it to the next version — so it needs deleting before the next release is published, or that release ships `0.7.2` binaries and the bug fires one last time.

shiv3:
Valid, and a worse bug than the one this PR set out to fix. Addressed in d4747c5.

I checked the pinned action's source (`1deb371`) rather than reasoning about it. `create-release.ts`:

```js
} else if (!isNewRelease && !release.draft) {
  // updateRelease changes the tags of draft releases, creating duplicate releases.
  // Therefore we only update published releases.
  console.log('Updating name and body of existing release...');
  await github.rest.repos.updateRelease({
    ..., body: bodyFileContent || body,
  });
}
```

That is exactly the path `tagName` + `releaseDraft: false` takes: the release for the tag exists and is published, so `isNewRelease` is false and `release.draft` is false. With `releaseBody` omitted, `body` is `""` — so it would blank release-drafter's notes, five times over, once per matrix leg. Trading a wrong-binaries bug for a wiped-notes bug isn't a fix.

I didn't pass the body through, though. `index.ts` only resolves a release when `tagName && !releaseId`, so handing it an **id** skips create, update and rename altogether and leaves tauri-action doing the one thing it's wanted for here — uploading assets. That's strictly narrower than round-tripping the body, and there's no window where five parallel jobs are racing to rewrite the same field.

So: a new `resolve-release` job resolves the id once before the matrix fans out, and the build step now passes `releaseId` instead of `tagName`/`releaseName`/`releaseBody`.

That job also subsumes your first-pass finding. If no release exists for the tag it creates one with `generate_release_notes` rather than failing the build — the direct-tag-push case ends up with real notes instead of an empty body, and the binaries still have somewhere to land.

Verified by executing the `resolve-release` script itself — parsed out of the workflow at test time so it can't drift from what ships — with `listReleases`/`createRelease` mocked:

| ref | release_id | createRelease | notes touched |
|---|---|---|---|
| tag, release exists (normal flow) | existing id | not called | no — neither create nor update runs |
| tag, no release (direct push) | newly created id | called, `generate_release_notes` | n/a, fresh release |
| branch (`workflow_dispatch`) | unset | not called | no — tauri-action skips uploads and just smoke-builds |

Issue #239, opened 2026-07-23
Add configurable network condition simulation


juherr:
The simulator could become much more useful for CSMS robustness testing by supporting configurable network conditions.

In production, charging stations often communicate through mobile networks, underground car parks or unstable local infrastructure. A simulator running over a perfect connection does not reproduce many real-world failures.

## Proposal

Add an optional network simulation layer that can be configured globally, per charger or per scenario run.

Possible conditions:

- fixed latency;
- random latency within a configured range;
- temporary connection loss;
- forced WebSocket disconnection;
- delayed reconnection;
- message drop probability;
- message duplication probability;
- response timeout simulation.

It would also be useful to apply these rules selectively:

- charger to CSMS;
- CSMS to charger;
- both directions;
- specific OCPP actions;
- specific scenario steps;
- only after a given event.

For random delays or message loss, supporting a random seed would make failures reproducible.

The UI could clearly indicate when network simulation is active and expose related events such as delayed, dropped or duplicated messages.

A first iteration limited to latency, forced disconnection and delayed reconnection would already cover many practical test cases.

shiv3:
Status: the first iteration proposed here shipped in #242 (follow-up fix in #243).

Implemented:

- latency rules with a fixed delay plus optional jitter, applied upstream / downstream / both, optionally matched to specific OCPP actions;
- forced disconnection with a configurable reconnect delay;
- periodic disconnection with interval jitter;
- a seeded RNG so randomized timing is reproducible;
- configuration globally or per charger (per-CP rules inherit and override the global layer), with a UI editor and a badge indicating when simulation is active.

Not implemented yet: message drop probability, message duplication, response-timeout simulation, and scoping rules to scenario steps or "only after a given event". Keeping this issue open for those.

PR #242, opened 2026-07-25
feat(network-sim): configurable network condition simulation (latency, forced disconnect, delayed reconnect)

shiv3:
Closes the first iteration asked for in #239: an in-process network-condition
simulation layer covering **latency, forced disconnection and delayed
reconnection**, configurable globally or per charge point, deterministic under a
seed, and drivable from the web console, the daemon RPC surface and MCP.

Refs #239 (Phase 1 only — drop/duplication, response-timeout faults and
scenario-step activation stay open for Phase 2).

## Summary

Network simulation is **off by default and bit-identical to today when
disabled** — no queue, no timers, no behavioural change on the send/receive
paths. When enabled, a per-charge-point controller sits inside `OCPPWebSocket`
and owns two FIFO delay pipelines (upstream/downstream) plus disconnect
scheduling:

- **Latency** — fixed delay + seeded jitter, filterable by direction and by
  OCPP action (CALL frames), additive across matching rules, saturating at 120 s.
- **Forced disconnection** — manual (triggered from the UI/RPC/MCP) or periodic,
  with an injected reconnect delay honoured as a lower bound on the first
  reconnect attempt.
- **Determinism** — `cyrb128` + `xoshiro128**` with per-charger and per-rule seed
  derivation, so the same seed reproduces the same faults; two charge points
  never perturb each other's sequences.

Configuration is a two-layer model (global + per-CP) where a per-CP `null` entry
is a tombstone removing an inherited rule, resolved per charge point into the
config the controller consumes. Rules apply to WebSocket charge points only;
SOAP CPs are unaffected and hide the UI.

## Design

The hard part was not the delay itself — it was making a delayed frame safe
inside OCPP's ordering and custody rules. Three mechanisms carry that weight:

- **Write settlement.** `sendAction`/`sendResult`/`sendError` keep their existing
  synchronous return, and gain a settlement callback that fires exactly once
  (`written`, or a typed drop). The 1.6 serializer now arms its 30 s response
  timeout and stamps heartbeat activity **at physical write**, not at enqueue, so
  a latency-delayed CALL holds the serialization slot without being "on the
  wire" — one outstanding CALL stays one outstanding CALL even at a 120 s delay.
  Custody is centralised in `salvageOrDiscard`, so a frame dropped before the
  wire still reaches `PendingMessageQueue` exactly like the pre-existing
  send-false path.
- **Close transaction + generations.** Every close path (network, operator,
  simulated, internal reset) runs one idempotent per-generation transaction:
  latch the cause, drain the pipelines synchronously, tear down handlers, then
  finalise and only then flush deferred response effects. A generation guard
  stops a stale socket's asynchronous close from draining or reconnecting the
  session that replaced it. Post-response side effects (Reset, TriggerMessage,
  firmware/diagnostics…) moved off bare `queueMicrotask` onto a shared
  `ResponseEffectQueue` that runs them at settlement and suppresses them for
  operator/internal/disposed closes — a queued `Reset.conf` drained by a user
  disconnect must not reboot the charge point.
- **Reconnect reservation.** The injected delay is an absolute deadline
  installed *before* the drain, owned by a single coalescing scheduler, surviving
  timer cancellation, and consumed only when a socket-open actually begins — so a
  Hard Reset triggered by the simulated loss cannot shorten the outage.

Two smaller correctness items came along for the ride: `pending_messages` gained
a monotonic `seq` column (millisecond `created_at` ties are now routine because
close-time salvage enqueues in bursts, and could reorder StartTransaction after
StopTransaction across a restart), and `reset` is unified onto one atomic
cause-aware operation across the daemon and remote clients instead of a
disconnect+connect pair.

Config validation is a trust boundary (RPC, kv, browser storage all feed it), so
it is strict reject-only over a single frozen snapshot: own-property reads,
accessor/symbol/bigint rejection, dense-array checks, and depth/node/byte budgets
that abort before a shared-reference structure can expand into an oversized
serialization.

## Surfaces

- **v3 console** — a global editor on `/settings`; a per-CP layered editor on the
  CP page showing each rule as inherited / overridden / disabled / local with
  override, revert, disable, delete and tri-state enable; a badge on the CP
  header and dashboard cards; one Force-disconnect button per manual rule.
- **Daemon** — five typed Socket.IO methods (`network_sim.global.get|save`,
  `network_sim.cp.get|save`, `network_sim.disconnect.trigger`) with typed trigger
  errors, three curated MCP tools, and `kv`-backed persistence that is restored
  and attached to a charge point *before* it auto-connects.
- **Local (browser) mode** — a versioned `networkSim` storage key applied to
  in-page charge points before connect and re-applied live on edit.
- **Observability** — a `NETWORK_SIM` log type across all four viewers, with
  wire-derived values control-character escaped so a hostile CSMS cannot forge a
  log line.

## Commits

27 commits in three phases: transport core (PRNG, config, policy, pipeline,
controller, `OCPPWebSocket`/`ChargePoint` integration, 1.6 and 2.x custody,
effects, `seq` migration, logging, harness + wire integration), daemon
(protocol schemas, `NetworkSimManager`, RPC dispatch, MCP, restart), UI
(local wiring, snapshot summary, editors, badges, composed tests + docs).

## Testing

- vitest: **1948 tests / 171 files green**; new suites for the PRNG vectors,
  validator (incl. prototype-pollution, accessor, byte-budget and
  shared-reference-DAG cases), pipeline ordering and settlement, controller,
  transport generations/close/reservation, both protocol handlers, the effect
  queue, the `seq` migration, and the console components.
- `bun test`: wire-level integration against the mock CSMS (extended with
  server-initiated close, per-connection transcripts and per-frame timestamps)
  covering latency lower bounds, forced disconnect + delayed reconnect,
  transaction salvage across a disconnect, stale downstream frames dropped on
  reconnect, and 1.6J / 2.0.1 / 2.1; daemon RPC, MCP, lifecycle and restart
  suites. Timing assertions are lower-bound + eventual-arrival only, so they do
  not depend on CI speed.
- `tsc -b --noEmit` adds no errors against a pre-change baseline; eslint is clean
  on every production file this branch touches.

## Summary by CodeRabbit

* **New Features**
  * Added configurable **Network Simulation** for **WebSocket** charge points (latency with jitter, manual disconnects, timed reconnect behavior) with **global** and **per-charge-point** layers, persisted and validated.
  * Introduced console UI for editing network simulation settings, including a status badge and per-rule manual disconnect controls.
  * Added runtime control via **socket RPC** and **MCP tools** (`network_sim.*`) plus updated server/client status wiring and documentation.

* **Bug Fixes**
  * Improved ordering and post-response behavior for charge-point message handling.
  * Confirmed **SOAP charge points are unaffected** by network simulation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,26 @@ ocpp-cp-sim --ws-url wss://steve.example.com/steve/websocket/CentralSystemServic
 
 Security extension configuration keys include `SecurityProfile`, `AuthorizationKey`, `AdditionalRootCertificateCheck`, `CertificateSignedMaxChainSize`, `CertificateStoreMaxLength`, and `CpoName`. The simulator can send `SecurityEventNotification` and `SignCertificate`, handle inbound `CertificateSigned`, and exposes JSON-mode RPC commands `security_event_notification` and `sign_certificate`.
 
+## Network Simulation
+
+Phase 1 of network condition simulation adds in-process fault injection for WebSocket charge points: configurable latency (with jitter), forced disconnects, and delayed reconnection. Configuration is **off by default** and applies globally (all CPs) or per-charger with null-tombstone override semantics. Same seed always produces the same fault sequence (deterministic via a seeded PRNG).
+
+**Configuration** — via the browser UI (`/v3/settings` for global, `/v3/cp/:id` for per-charger) or the daemon's RPC/MCP control plane. No startup flags in Phase 1 — configuration is applied at runtime or persisted to the state database (`networkSim:global`, `networkSim:cp:<cpId>` keys).
+
+**Layering & null semantics** — global and per-CP layers merge with per-CP overrides winning. Setting per-CP config to `null` deletes the override and reverts to the global layer. A `null` global and `null` per-CP means the feature is disabled.
+
+**Determinism** — the rule engine is seeded; identical seed + rule order produces identical fault timings across different instances. Useful for reproducing issues or running A/B tests.
+
+**WebSocket only** — OCPP 1.2/1.5/1.6 (SOAP) charge points are unaffected. The simulation applies only to WebSocket transports.
+
+**Documented limitations** (these require future proxy-mode work):
+
+- Cannot reproduce TCP-level half-open connection states.
+- TLS-level handshake failures are out of scope.
+- Browser Ping/Pong frames and their failures are not simulated.
+
+For configuration details, rule schemas, and protocol-timer behavior (e.g., how a delayed BootNotification affects the boot acceptance gate), see [docs/server.md § Network Simulation](docs/server.md#network-simulation).
+
 ## AI Agent & Automation Testing
 
 The daemon exposes a single Socket.IO control connection and emits structured logs, making it a scriptable OCPP stub that any AI agent or test harness can drive.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -239,6 +239,8 @@ There is no Unix-domain socket listener. `--unix-socket <path|none>` is still
 accepted for launcher compatibility, prints a deprecation warning, and is
 ignored.
 
+**Network Simulation** — configuration is applied at runtime via the daemon's Socket.IO RPC methods or MCP tools (see [docs/server.md § Network Simulation](server.md#network-simulation)); there are no `--network-sim-*` startup flags. In Phase 1, configure latency, forced disconnects, and delayed reconnection through the browser UI (`/v3/settings` for global, `/v3/cp/:id` for per-CP), or directly via RPC methods `network_sim.global.save`, `network_sim.cp.save`, and `network_sim.disconnect.trigger`.
+
 #### Multiple Charge Points in One Process
 
 `--cp-id` becomes optional in server mode. Additional CPs can be added at
--- a/docs/server.md
+++ b/docs/server.md
@@ -162,6 +162,18 @@ Runtime packages for this control plane are `socket.io`,
 | `events.subscribe`   | `{ "scope": "*" \| "registry" \| "<cpId>" }`                                                                                                                                                          | Join event rooms and return an atomic snapshot.                                       |
 | `events.unsubscribe` | `{ "scope": "*" \| "registry" \| "<cpId>" }`                                                                                                                                                          | Leave an event room.                                                                  |
 
+### Network Simulation methods
+
+| Method                           | Params                                             | Result / purpose                                                                                                                                                        |
+| -------------------------------- | -------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `network_sim.global.get`         | `{}`                                               | Return the global network-sim layer config, or `null` if disabled.                                                                                                      |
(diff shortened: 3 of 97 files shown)


shiv3:
All review findings from both passes are fixed. Summary of what changed and why:

**Production bugs**

- `writeUpstreamPhysical` returning silently on a closed socket (`f62fc29`) — confirmed. A frame held in the latency pipeline can be released after the socket drops without an `onclose`, so no close transaction drains the pipeline; the frame settled `written` and its transaction CALL left `PendingMessageQueue` custody without reaching the wire. It now throws, the pipeline maps that to `write_failed`, and `onSerialSettled` salvages it. Regression test asserts `write_failed` and fails with `written` against the old code.
- `PendingMessageQueue` counter conflation (`f62fc29`) — confirmed. `nextSeq` and the message-id suffix were one counter, so `seq` advanced by two per enqueue. Split them; the id counter is seeded from the same high-water mark on load so a queue reopened inside one millisecond can't collide on `(cp_id, message_id)`.
- `seq` leaking through `PendingMessage` (`874475e`) — confirmed. `all()`, `dequeue()` and the `flush()` callback stripped only `messageId`. All three now go through one projection that drops both storage columns.

**UI** (`653b437`) — all six confirmed: errors cleared on every form mutation, effect keyed on prop *contents* rather than identity (`CpDetailPage` rebuilds `inheritedRules` inline on every snapshot poll, so identity-keyed deps wiped in-progress edits — covered by a DOM test), per-CP error keys use the original row index, the hard-coded `seed: 1` that shadowed the derived per-charger seed is gone, `SettingsPage` surfaces a rejected `getNetworkSimGlobal`, and `LocalChargePointService` throws on an invalid config so local and remote agree.

**Test defects** (`30b72fb`, `874475e`) — the settlement suites asserted constants or defined the policy inline in their own callback and then checked that callback ran; neither touched production code. They now install a handler returning a `HandlerOutcome`, deliver an inbound CALL and settle the response the handler actually sent — removing the `ResponseEffectQueue.register` call from `OCPPMessageHandler` fails two of them. Also: the unreachable restart-ordering assertion is now unconditional, `mockCsms` servers are torn down in `afterEach` (and `stop()` is idempotent and bounded — Bun 1.3 frees the listener immediately but never settles `stop()`'s promise once the server has closed a WebSocket itself, which wedged the new cleanup), the in-place `frames.reverse()` sites are copies, and the `GetLog` effect test covers the scheduled `Uploaded` notification.

**One regression the review didn't reach**, found while chasing the red `test` job (`c781059`): `v16RegistryDispatch` cast handler results straight to `SoapPayload`, so once the 1.6 CALL handlers moved to `HandlerOutcome`, every SOAP charge point answered `TriggerMessage`/`Reset`/`GetDiagnostics`/`GetLog`/firmware with the wrapper object instead of the response, and the post-response effect never ran. Bisected to the migration commit — passes on `main`, failed from there onward. Fixed by unwrapping and deferring the effect past the reply.

CI is green (vitest 1958, bun 292). Note that CodeRabbit reported **"Review rate limited"** on the last three pushes, so `f62fc29..c781059` has not actually been reviewed.

PR #253, opened 2026-07-30
fix(scenario): re-arm connect-triggered scenarios after a reconnect

shiv3:
## Problem

A `triggerOn: "connect"` scenario is armed once per connect, recorded on `Connector.lastAutoStartedScenarioKey` so that re-emitting `Available` doesn't restart something already running. That arm survived the socket dying, so the scenario **never re-fired after a reconnect**.

Found running v0.7.5 as a daemon on a dev cluster with `--scenario-template essential-cp-behavior --scenario-connector all` over 7 connectors:

1. All 7 runs park on the `remoteStartTrigger` node.
2. A network-sim `manual-disconnect` (code 1000, clean close) fails every one with `Scenario execution failed: Disconnected while waiting for remote start`.
3. The WebSocket reconnects and BootNotification is accepted — but no scenario restarts.
4. The simulator silently stops answering `RemoteStartTransaction` until an operator re-fires `run_scenario_template` on every connector by hand.

For a deployed simulator that means E2E traffic breaks quietly on every disconnect.

The arm key is also persisted (`connector_runtime.last_auto_started_scenario_key`), so a stale value can outlive a process restart.

## Fix

Clear the arm in `ChargePoint.teardownAfterClose`, **before** the `status = Unavailable` assignment: that setter cascades Unavailable onto every connector, and the resulting `statusChange` is what drives `persistConnectorRuntime`. Clearing afterwards would persist the stale key.

Fixing it in the domain covers both auto-start engines at once — the daemon's `CLIChargePointService.tryAutoStartForConnector` and browser local mode's `useEffect` in `Connector.tsx`, which had the identical bug.

No new flag: the CLI bootstrap is covered automatically.

Also prefixes scenario log lines with the bound connector. The executor's only prefix is the scenario *name*, which a per-connector template instantiation shares, so the disconnect emitted 7 byte-identical failure lines with nothing to tell them apart.

## Behaviour change

A **completed** `triggerOn: "connect"` scenario now runs again on each reconnect. That's the literal reading of the trigger, and it's what keeps a long-lived simulator responsive. A status oscillation on a still-connected CP does **not** re-fire. Documented in `docs/scenario-format.md` with a pointer to explicit `run_scenario` for run-once-per-process cases.

## Tests

- `ChargePoint.scenarioRearm.bun.test.ts` — the arm is cleared on disconnect, cleared *before* the Unavailable cascade (so the persisted snapshot is clean), and left alone while connected.
- `scenarioRearmOnReconnect.bun.test.ts` — end to end over a real socket against a mock CSMS: connect → boot → run → disconnect → reconnect → boot → a **second** run (watched via `scenario_report`'s `runId`). Verified this fails without the fix.
- `scenarioLogConnectorPrefix.bun.test.ts` — same-named runs on different connectors produce distinguishable lines.

Full suites green: 320 bun tests, 2117 vitest tests. `tsc -b --noEmit` error count unchanged at 510 (all pre-existing).

First of three PRs from the same v0.7.5 field-verification pass (P1 → P2 API contract → P3 operational).

## Summary by CodeRabbit

* **Bug Fixes**
  * Connect-triggered scenarios now properly re-arm after disconnect/reconnect and don’t get stuck in an armed state.
  * Auto-start “arm” state is reliably cleared on teardown/disconnect, and persisted correctly when the socket drops.
  * Scenario logs are now prefixed with the connector identifier for clearer troubleshooting.

* **Documentation**
  * Expanded guidance for `triggerOn: "connect"` semantics, including reconnect behavior and when to use `run_scenario`.

* **Tests**
  * Added end-to-end and unit coverage for reconnect re-arming and disconnect-clearing/persistence behavior.

### DIFF
--- a/docs/scenario-format.md
+++ b/docs/scenario-format.md
@@ -108,6 +108,24 @@ shape depends on `type`).
 `Preparing`, `Charging`, `SuspendedEVSE`, `SuspendedEV`, `Finishing`,
 `Reserved`, `Unavailable`, `Faulted`.
 
+### `start` notes: `triggerOn: "connect"` fires on _every_ connect
+
+A manual-trigger scenario whose start node uses `triggerOn: "connect"` (the
+default) auto-starts once the charge point reaches `Available` after
+BootNotification is accepted. It is armed **per connection**: losing the
+WebSocket disarms it, so the scenario starts again on the next connect — even
+if the previous run had already finished.
+
+This matters for a long-lived simulator whose scenario answers CSMS-initiated
+calls (a `remoteStartTrigger` responder, say). The run dies with
+`Disconnected while waiting for remote start` when the socket drops, and
+re-arming is what makes the simulator answer again after it reconnects rather
+than going quietly unresponsive.
+
+If a scenario must run only once per process, drive it with an explicit
+`run_scenario` instead of `triggerOn: "connect"`. A status oscillation on a
+still-connected charge point does **not** re-fire the scenario.
+
 ### `responseOverride` notes
 
 Which `status` values are valid depends on `action` (e.g. `action:
--- a/src/cli/server/__tests__/scenarioRearmOnReconnect.bun.test.ts
+++ b/src/cli/server/__tests__/scenarioRearmOnReconnect.bun.test.ts
@@ -1,0 +1,197 @@
+/* eslint-disable @typescript-eslint/no-explicit-any -- ack payloads are loosely typed in tests */
+import { afterEach, describe, expect, it } from "bun:test";
+import type { Socket } from "socket.io-client";
+
+import {
+  connectTestClient,
+  startTestServer,
+  type TestServer,
+} from "./socketHarness";
+import { startMockCsms } from "../../../cp/infrastructure/transport/__tests__/mockCsms";
+
+/**
+ * A `triggerOn: "connect"` scenario that was killed by a WebSocket drop must
+ * come back on the next connect.
+ *
+ * Field report (v0.7.5, GKE dev): `essential-cp-behavior` bootstrapped over all
+ * connectors and parked on its remoteStartTrigger node died on a network-sim
+ * manual-disconnect with "Disconnected while waiting for remote start". The CP
+ * reconnected and BootNotification was accepted, but no scenario restarted --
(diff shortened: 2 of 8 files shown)


shiv3:
Valid — fixed in 359e2a6.

Verified the mechanism: `status = Unavailable` skips `connector.transaction !== null` (`ChargePoint.ts:1379`), and `persistConnectorRuntime` is driven purely by connector-level events (`service.ts:1791+`), so a connector that was charging when the socket dropped never wrote the cleared arm.

Impact is slightly wider than "a restart can restore the stale arm": it reopens the same silent failure through the restart path for any scenario with an id that is stable across restarts — anything loaded by id, or rehydrated by `restoreScenariosFromDatabase`. The arm key embeds the scenario id, so a restored stale key means the connect-triggered scenario never re-arms. The CLI `--scenario-template` bootstrap escaped it only incidentally, because it mints a fresh instance id every boot and that changes the key.

Fix persists every connector's snapshot on the `disconnected` event rather than relying on which connectors happened to emit a statusChange. `teardownAfterClose` has already run by then — the WebSocket `onclose` path calls it before the emit (`ChargePoint.ts:882`, `887`) — so the persisted value is the cleared one.

Regression test `scenarioRearmPersistsOnDisconnect.bun.test.ts` drives the active-transaction case end to end against a mock CSMS: connect → auto-start → Authorize → StartTransaction → open transaction → disconnect, then reads `connector_runtime` directly. It asserts the pre-condition that the arm was persisted while running, and pins the branch under test (`status != Unavailable`, `transaction_json != null`) so it can't pass for the wrong reason if that cascade behaviour ever changes. Confirmed failing before the fix.

Full suites green: 320 bun, 2117 vitest. `tsc -b --noEmit` unchanged at 510.

shiv3:
Valid — fixed in 48ea576. Adopted the polling shape; a fixed sleep is a flake source and the rest of the suite already uses condition-based waits.

Extracted the row read into a `readRow()` helper so the poll and the final assertion query the same thing, and kept the bound at 10s. Verified the bound still does its job: with the persistence call removed, the test fails after ~10s rather than passing — so this doesn't turn the assertion into a no-op.

Issue #214, opened 2026-07-14
Feature: Publish a versioned JSON Schema for scenario definitions


juherr:
## Background

During the discussion on PR #210, it was suggested that the scenario format should be described by a versioned JSON Schema instead of being implicitly defined by the TypeScript interfaces alone.

Reference:
https://github.com/shiv3/ocpp-cp-simulator/pull/210#issuecomment-4966155984

## Motivation

Today, the scenario format is primarily described by the TypeScript model (`ScenarioDefinition`) and validated through runtime code.

Publishing an official JSON Schema would make the format a stable, implementation-independent contract.

This would benefit both simulator users and external tooling.

## Benefits

- validate scenario files before loading or executing them;
- IDE autocompletion and documentation;
- easier authoring of scenarios outside the simulator UI;
- compatibility checks in CI;
- explicit versioning of the external format;
- simpler interoperability with third-party tools.

Most importantly, it would provide the foundation for higher-level APIs and DSLs.

For example:

- Java/Kotlin fluent builders;
- TypeScript builders;
- Go libraries;
- Python tooling;
- AI-assisted scenario generation.

Instead of manually constructing JSON objects, these libraries could generate scenario files that are validated against the same schema used by the simulator.

## Possible implementation

- publish a JSON Schema (Draft 2020-12);
- introduce a `schemaVersion` field in scenario documents;
- validate imported scenarios against the schema before execution;
- generate the TypeScript types from the schema (or vice versa) to avoid duplication;
- include the schema in releases so external tools can depend on a versioned contract.

## Future opportunities

Once a stable schema exists, it becomes possible to build:

- language-specific DSLs;
- scenario converters;
- linters;
- visual editors;
- code generators;
- migration tools between schema versions.

This would make the scenario format much easier to adopt across the OCPP ecosystem while keeping the simulator as the reference implementation.

PR #254, opened 2026-07-30
fix(api): validate loaded scenarios, tail logs.get, keep terminal scenario status

shiv3:
Three API-contract problems found verifying v0.7.5 over RPC and MCP. Second of three PRs from that pass (P1 was #253; P3 operational follows).

## `load_scenario` silently accepted an incomplete definition

`loadScenario` keyed its runtime map on `definition.id` and returned it unchecked. A payload with no `id`:

- was stored under the key `undefined`,
- answered `{}` instead of `{ scenarioId }`,
- left an entry in `list_scenarios` with a name but no id — neither runnable nor removable,
- and got persisted to `state.db`.

Gated in two places, on the fields the runtime keys on (the schema's own `required` set: `id`, `name`, `targetType`, `nodes`, `edges`):

1. The `load_scenario` params schema, so the wire answers `invalid_params` for Socket.IO and MCP alike.
2. `loadScenario` itself, covering the `file` parameter and the startup loaders that never see that schema.

Full-schema conformance stays advisory (#214) — only this minimal set is a hard gate. `syncConnectorRuntimeScenarios` *skips* an unloadable definition instead of throwing: it reconciles a set that is already the DB's truth, and one bad entry must not fail the rest.

> The reported "`list_scenarios` has no `scenarioId`" was this same bug: `listScenarios` already returned the field, and the id-less entry serialised without its `undefined` value.

## `logs.get {limit}` returned the oldest N

`entries.slice(0, limit)` over an `ORDER BY id ASC` query. On a charge point that had been up for days, no `limit` could reach recent activity — the one thing the parameter is for (verification had to fall back to `kubectl logs`).

| Params | Returns |
| --- | --- |
| `{limit: 100}` | the newest 100, oldest-first within the window |
| `{limit: 100, offset: 100}` | the 100 before those — `offset` pages back in time |
| `{limit: 100, order: "desc"}` | the newest 100, newest-first |

**This changes the default for an existing parameter.** `docs/server.md` documents the migration, and the MCP `get_logs` description spells the semantics out so an agent doesn't have to guess.

## `scenario_status` went null once a run ended

It read only the live-executor map that `runScenario`'s `finally()` deletes from. A client polling for `state === "completed"` never succeeded, and could not distinguish that from an unknown `scenarioId` — while the run was still sitting in `scenario_report`.

Terminal status now survives until the scenario is removed. `state` is taken from the run's `executionState`, so `scenario_status` and `scenario_report` can never disagree about how a run ended. `null` now means exactly "no run on record for this id", which is what makes the two cases distinguishable. `list_scenarios` gains `state` and `mode` from the same source (pure additions; local/browser mode reports live state only, since its manager keeps no terminal context).

## Tests

- `service.loadScenarioValidation.bun.test.ts` — each required field rejected, empty/whitespace id, bad `targetType`, non-array `nodes`, non-object payloads (previously a `TypeError`), and nothing half-loaded.
- `logWindow.bun.test.ts` — tail, paging, `desc`, boundaries, no input mutation. Verified these fail against the old head-slice.
- `service.terminalScenarioStatus.bun.test.ts` — terminal state after completion and after a manual stop, agreement with the report, cleared on remove, replaced on re-run, `list_scenarios` state/mode.
- `scenarioApiContract.bun.test.ts` — all three over the real socket: `invalid_params` acks, log tail/paging/order against the actual buffer, and a poll loop that used to spin until timeout.
- `bounds.test.ts` fixtures are now well-shaped so the size-bound tests can only fail on size, plus a case for the new shape gate.

Full suites green: 344 bun tests, 2118 vitest tests. `tsc -b --noEmit` error count back to the 510 baseline (the tighter param type surfaced one real cast in `RemoteChargePointService`, now typed against the method's params).

## Backward compatibility

| Change | Compatible? |
| --- | --- |
| `load_scenario` required-field rejection | **No** — an incomplete definition now fails instead of loading unusably |
| `logs.get` `limit` default | **No** — newest N instead of oldest N |
| terminal `scenario_status` | Yes — was null, now a terminal object; unknown ids stay null |
| `list_scenarios` `state`/`mode` | Yes — added fields |
| `logs.get` `offset`/`order` | Yes — new optional params |


shiv3:
@coderabbitai review

PR #255, opened 2026-07-30
fix(daemon): make template loads idempotent and report the real version

shiv3:
Two operational problems found verifying v0.7.5. Third of three PRs from that pass (P1 was #253, API contract #254).

## Duplicate scenarios accumulate in `state.db`

Template instances get a `${templateId}-${cpId}-c${connectorId}-${Date.now()}-${suffix}` id, are persisted, and are all rehydrated on the next boot — so every pod restart and every template re-run added one more copy per connector. The dev cluster had three `Essential CP Behavior` entries with weeks-old timestamps stacked on one connector, and nothing ever cleaned them up.

Instantiating a template is now **idempotent per (template, connector)**: the earlier instance of that template on that connector is dropped from memory and the DB first.

- A different template on the same connector is untouched.
- The same template on another connector is untouched.
- `createScenario` is unchanged, so loading a template twice from the editor still yields two independent instances.

Matching has to answer "another instance of the same template", which a timestamped id can't without parsing — so `ScenarioDefinition` gains an optional `templateId`, stamped by `templateFromJson`. Rows persisted before that field existed are matched on the exact id format `createScenario` produces, so **an existing `state.db` tidies itself on the next boot** — the reported dev-env pile-up included. Deliberately *not* a bare `id.startsWith(templateId)`: that would also delete a different template whose id happens to start with this one (there's a test for it).

The schema root is `additionalProperties: true`, so the new field breaks no contract. I folded it into schema v1.0 alongside `strictCompatibility` rather than bumping to 1.1 — both are additive and optional, and this matches how `strictCompatibility` was added in #250. Happy to bump the version instead if you'd rather follow the doc's minor-bump rule strictly.

## Release builds reported version `0.0.0`

`scenario_report.simulatorVersion` and MCP `serverInfo.version` were both hard-coded literals matching package.json's unstamped placeholder, and `/v1/healthz` carried no version at all. A report from a GHCR release was indistinguishable from one off a dev checkout.

**No build-pipeline change is needed.** The release workflows already run `npm version --no-git-tag-version` before building (`docker-publish.yml`, `release.yml`) and the runtime stage does `COPY package.json` — so a released artifact's package.json already carries the real semver. The CLI runs from source under bun, so there's no bundler `define` to hook; read it at runtime instead.

`appVersion()` resolves in order:

1. `APP_VERSION`, for callers that build an image without going through `npm version`
2. package.json's `version`, unless it's still the `0.0.0` placeholder
3. `0.0.0-dev` — an unstamped build says so, rather than impersonating release 0.0.0

`/v1/healthz` gains `version` alongside the unchanged `ok` field that probes assert on, so you can confirm which image a deployment is serving without an authenticated round-trip.

## Tests

- `service.templateIdempotency.bun.test.ts` — replacement, other-template and other-connector isolation, `templateId` stamping, repeated `seedDefaultScenarios`, DB-level cleanup verified through a restart (fresh service over the same DB), legacy rows without `templateId`, and the prefix-lookalike case.
- `appVersion.bun.test.ts` — precedence, trimming, per-call env read, dev marker.
- `versionSurfaces.bun.test.ts` — the stamp reaches `/v1/healthz` and `scenario_report.simulatorVersion`, and `ok` still comes back untouched.

Three existing healthz tests asserted the exact response body with `toEqual`. Their invariant is that healthz answers unauthenticated, so they now match on `ok` rather than the whole object.

Full suites green: 328 bun tests, 2117 vitest tests. `tsc -b --noEmit` error count unchanged at the 510 baseline.

> One note on the vitest run: a first pass showed 4 failures in `useChargePoints.registry`, `network-sim/config`, and `DataProvider.detect`. All pass in isolation and on a clean re-run of the full suite — they're load-sensitive timeouts (the network-sim one uses ~1.9s of a 2s budget), not caused by this change.

## Backward compatibility

| Change | Compatible? |
| --- | --- |
| Template load replaces the prior instance | Yes — same ids and behaviour per boot, minus the accumulation. A running instance of the *same* template on that connector is stopped, which is what "replace" means. |
| `templateId` on `ScenarioDefinition` | Yes — optional, `additionalProperties: true` |
| Real version in reports / MCP / healthz | Yes — `0.0.0` was a placeholder; `ok` on healthz is unchanged |


shiv3:
@coderabbitai review

Issue #257, opened 2026-07-30
Incomplete CS→CP SOAP coverage: OCPP 1.5 implements only Reset (1/15), DataTransfer missing in 1.5 and 1.6S


juherr:
## Summary

The daemon answers CS→CP SOAP calls unevenly across dialects. OCPP **1.2** is complete, **1.6S** is one operation short, and **1.5** implements a single operation out of fifteen — even though `OCPP15_DIALECT` declares the full surface.

Everything below was measured against `0.7.5` by sending a well-formed WS-Addressing envelope for every operation the **official OCA `ChargePointService` WSDLs** define, and classifying the answer. Nothing is inferred from reading the code.

## Reference: what the official WSDLs define

| Dialect | Namespace | CS→CP operations |
|---|---|---|
| 1.2 | `urn://Ocpp/Cp/2010/08/` | 9 |
| 1.5 | `urn://Ocpp/Cp/2012/06/` | 15 |
| 1.6 | `urn://Ocpp/Cp/2015/10/` | 19 |

## Measured coverage

| Operation | 1.2 | 1.5 | 1.6S |
|---|:---:|:---:|:---:|
| ChangeAvailability | ✅ | ❌ | ✅ |
| ChangeConfiguration | ✅ | ❌ | ✅ |
| ClearCache | ✅ | ❌ | ✅ |
| GetDiagnostics | ✅ | ❌ | ✅ |
| RemoteStartTransaction | ✅ | ❌ | ✅ |
| RemoteStopTransaction | ✅ | ❌ | ✅ |
| Reset | ✅ | ✅ | ✅ |
| UnlockConnector | ✅ | ❌ | ✅ |
| UpdateFirmware | ✅ | ❌ | ✅ |
| CancelReservation | — | ❌ | ✅ |
| DataTransfer | — | ❌ | ❌ |
| GetConfiguration | — | ❌ | ✅ |
| GetLocalListVersion | — | ❌ | ✅ |
| ReserveNow | — | ❌ | ✅ |
| SendLocalList | — | ❌ | ✅ |
| ClearChargingProfile | — | — | ✅ |
| GetCompositeSchedule | — | — | ✅ |
| SetChargingProfile | — | — | ✅ |
| TriggerMessage | — | — | ✅ |
| **Total** | **9 / 9** | **1 / 15** | **18 / 19** |

[...]

Issue #178, opened 2026-07-11
v0.7.0 testing feedback: SOAP, security, logs, scenario UI and documentation


juherr:
# v0.7.0 testing feedback

I tested the `v0.7.0` release mainly against SteVe, using both OCPP-J and OCPP-S configurations.

Overall, this release adds several valuable capabilities, especially SOAP support, security profiles, scenarios, and improved OCPP version handling.

Below is a consolidated list of the issues and improvement opportunities found during testing.

## Test environment

- Simulator version: `v0.7.0`
- Reference CSMS: SteVe `3.13.0`
- SteVe database version: `1.1.4`
- Protocols tested:
  - OCPP-J
  - OCPP-S `1.5`

SteVe endpoint patterns used during testing:

- JSON/WebSocket:
  `ws://<server>/steve/websocket/CentralSystemService/`
- SOAP:
  `http://<server>/steve/services/CentralSystemService`

SOAP simulator configuration used for the failing test:

- Central System URL:
  `http://<server>/steve/services/CentralSystemService`
- SOAP Callback URL:
  `https://<simulator>/ocpp/soap/<Charge Point ID>`
- SOAP Path:
  empty

---

### 1.1 Duplicate Basic Authentication settings

Basic Authentication can currently be configured in two different places:

- `Optional Settings > Enable Basic Authentication`
- `Security > Security Profile > Basic Auth`

This creates ambiguity about which setting is authoritative and how the two settings interact.

The legacy setting under `Optional Settings` should preferably be removed, and authentication should be managed from the Security section.

Backward compatibility must nevertheless be considered for OCPP versions older than 1.6, because the OCPP 1.6 security-profile model may not directly apply to these versions.

Possible approaches:

- Keep a generic authentication setting for pre-1.6 versions only.
- Display security profiles only for versions supporting them.
- Migrate the legacy setting internally while exposing a single consistent UI.

### 1.2 URL scheme should follow the selected security level

Changing the security profile should automatically update the URL scheme:

| Transport | Without TLS | With TLS |
|---|---|---|
| OCPP-J | `ws://` | `wss://` |
| OCPP-S | `http://` | `https://` |

At minimum, the UI should warn when the selected URL scheme is inconsistent with the configured security profile.

---

### 2.1 Horizontal scrolling

The logs cannot be scrolled horizontally.

This becomes particularly problematic when the Scenario panel is open because the remaining log area is too narrow to display complete messages.

[...]

PR #258, opened 2026-07-30
fix(soap): drive the SOAP ChargePointService from SteVe — 1.5 + DataTransfer dispatch, optional wsa:From (#256, #257)

### LINKED ISSUE #256: SOAP ChargePointService rejects CSMS requests that omit wsa:From (blocks SteVe / Apache CXF)
## Summary

The daemon's SOAP `ChargePointService` endpoint requires the WS-Addressing `<wsa:From>` header on inbound requests. Apache CXF — and therefore [SteVe](https://github.com/steve-community/steve), the reference open-source CSMS — does not send it on CS→CP calls. The result is that **no** CSMS→CP operation can reach the simulator from SteVe, in any SOAP dialect (1.2, 1.5, 1.6S).

## Reproduce

Charge point registered with `ocppVersion: OCPP-1.6S`, `soapCallbackUrl: http://simulator-ocpp:9700/ocpp/soap/TEST_16S/ChargePointService`. Boot works — SteVe stores `ocpp1.6S` and the endpoint address correctly, heartbeats flow.

Trigger any operation from SteVe (*Operations → OCPP v1.6 → Get Configuration*). SteVe's task detail reports:

```
org.apache.cxf.transport.http.HTTPException: HTTP response '400: (POST http://simulator-ocpp:9700/ocpp/soap/TEST_16S/ChargePointService)
```

Request captured on the wire with `tcpdump` (verbatim, CXF 4.2.2):

```xml
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope">
  <soap:Header>
    <Action xmlns="http://www.w3.org/2005/08/addressing">/GetConfiguration</Action>
    <MessageID xmlns="http://www.w3.org/2005/08/addressing">urn:uuid:d7eaf8bf-b8ef-4851-aa11-f9818d568dd2</MessageID>
    <To xmlns="http://www.w3.org/2005/08/addressing">http://simulator-ocpp:9700/ocpp/soap/TEST_16S/ChargePointService</To>
    <ReplyTo xmlns="http://www.w3.org/2005/08/addressing"><Address>http://www.w3.org/2005/08/addressing/anonymous</Address></ReplyTo>
    <chargeBoxIdentity xmlns="urn://Ocpp/Cp/2015/10/">TEST_16S</chargeBoxIdentity>
  </soap:Header>
  <soap:Body><getConfigurationRequest xmlns="urn://Ocpp/Cp/2015/10/"/></soap:Body>
</soap:Envelope>
```

Response:

```xml
<s:Envelope xmlns:s="http://www.w3.org/2003/05/soap-envelope"><s:Body><s:Fault>
  <s:Code><s:Value>s:Sender</s:Value></s:Code>
  <s:Reason><s:Text xml:lang="en">SOAP envelope is missing From</s:Text></s:Reason>
</s:Fault></s:Body></s:Envelope>
```

Replaying that exact envelope with a single `<From><Address>…</Address></From>` added returns **200** and a complete `getConfigurationResponse`. Same in 1.5 with `Reset` → `resetResponse Accepted`. `From` is the only difference.

## What the specs say

Stating the strongest counter-argument first, because it is real: **OCPP-S 1.6 §3.7 does say "Required: Yes"** —

> **3.7. From** — **Required: Yes**
>
> According to the WS-A 1.0 Core specification the from-field is marked as optional. However, for OCPP the field MUST be filled in. It is the only way for a Central System to know how to address server-initiated messages to a Charge Point. The address specified must be the address where the sender listens for incoming SOAP requests.

("Open Charge Point Protocol SOAP 1.6, OCPP-S 1.6 Specification", FINAL, 2015-10-08.)

So the current behaviour is not baseless. But four things argue it should not apply to a **CS→CP** request:

1. **The stated rationale is CP→CS only** — "the only way for a *Central System* to know how to address server-initiated messages to a *Charge Point*". On a CS→CP request, `From` would carry the Central System's address, which OCPP never consumes: the CP already knows its CS from configuration.
2. **The document scopes by direction when it means to, one section earlier.** §3.6: *"**Required: Yes, for responses only.** RelatesTo should not be filled in for request messages."* §3.7 carries no such qualifier — but it also states no obligation on the Central System.
3. **§5 restates the same rule as a CP-side `SHOULD`**: *"the Charge Point **SHOULD** set its URL in the WS-Addressing "From" field when sending a request"* — contradicting §3.7's `MUST`.
4. **§3 opens with a permissive default**: *"Anything that we do not specify in this chapter is optional or not necessary, in compliance with the WS-A specification."* The chapter's purpose was to tighten the CP→CS wire format.

**OCPP 1.5 has no such requirement at all.** The 1.5 specification has no SOAP-header chapter — its transport binding (ch. 8) covers only Charge Box Identity, Fault Response and Mobile Networks. The only two mentions of `From` in the whole document are in the mobile-networks section, and both concern the *Charge Box* setting its own URL when sending. For the 1.5 dialect the requirement has no basis whatsoever.

**The machine-readable normative artifacts and W3C agree:**

- The official `ChargePointService` WSDLs — [1.5 final](https://github.com/steve-community/ocpp-jaxb/blob/master/ocpp-jaxb/src/main/resources/wsdl/ocpp_chargepointservice_1.5_final.wsdl) and [1.6](https://github.com/steve-community/ocpp-jaxb/blob/master/ocpp-jaxb/src/main/resources/wsdl/OCPP_ChargePointService_1.6.wsdl) (same repo this project already cites in `dialect.ts` for field ordering) — declare exactly **one** `soap12:header`, `chargeBoxIdentity`. No `wsa:From` anywhere, and the WS-Policy is a bare `<wsaw:UsingAddressing />` with **no `wsdl:required`**.
- [WS-Addressing 1.0 — Core](https://www.w3.org/TR/ws-addr-core/) §3.1 gives `[source endpoint]` cardinality **`0..1`**; §3.2 calls it an "OPTIONAL element", against `wsa:Action`'s "REQUIRED element".
- [WS-Addressing 1.0 — WSDL Binding](https://www.w3.org/TR/2006/CR-ws-addr-wsdl-20060529/) §5.1.2 Table 5-2, request message in the request-response MEP that OCPP-S uses: `[source endpoint]` → Mandatory **N**, *"Unused in this MEP"*. The mandatory set is `[destination]`, `[action]`, `[reply endpoint]`, `[message id]` — exactly what CXF sends.

**And no OCPP-S implementation does this.** Surveyed: SteVe, motown-io/motown, voltbras/ts-ocpp, voltbras/go-ocpp, ShellRechargeSolutionsEU/ocpp-soap, IZIVIA/ocpp-toolkit — none sends `From` CS→CP, none faults on its absence. SteVe's tolerance is deliberate (`MessageHeaderInterceptor`: *"the absence of the field is not a deal breaker anymore"*), and the omission itself is stock Apache CXF: `MAPAggregatorImpl` sets `MessageID`/`Action`/`To`/`ReplyTo`/`FaultTo`/`RelatesTo` and never `From`. Python's `zeep` behaves the same way.

Two further notes:

- With `ReplyTo` = `…/anonymous` the response travels back on the same HTTP connection, so `From` carries no routing information the simulator needs here.
- The simulator already treats `From` as optional on *responses* (`optionalAddressHeader`); only the request path requires it. It also requires `ReplyTo`, which CXF happens to send — but that is optional per §3.8 too.

## Suggested fix

The requirement looks like it was correctly implemented for the CP→CS direction — the simulator does emit `From` on its own outbound requests (`soapEnvelope.ts:387-391`) — and then generalised to `kind === "request"`, which also catches inbound `ChargePointService` calls.

`src/cp/infrastructure/transport/soap/soapEnvelope.ts:813-821`:

```ts
const from =
  kind === "request"
    ? requireAddressHeader(header, "From")   // → optionalAddressHeader
    : optionalAddressHeader(header, "From");
```

If you would rather keep a strict reading of OCPP-S 1.6 §3.7, an opt-in flag (`--strict-soap-headers`) would work too — but please don't fault by default, since that makes the simulator unusable against SteVe and any other stock-SOAP CSMS. For the 1.5 dialect the requirement should go regardless: that spec has no header chapter at all.

Happy to open a PR if you'd like.

## Environment

- `ghcr.io/shiv3/ocpp-cp-simulator:0.7.5`, daemon mode, `--state-db`, behind a reverse proxy
- SteVe `3.14.0` (Apache CXF 4.2.2), same Docker network, direct container-to-container
- Reproduced on OCPP 1.2, 1.5 and 1.6S

shiv3:
## Summary

Fixes #256 and #257. As reported, the daemon's SOAP `ChargePointService` could not receive **any** CS→CP call from SteVe (Apache CXF), and its dispatch coverage was uneven across dialects (1.2 = 9/9, 1.6S = 18/19, 1.5 = 1/15). This makes CS→CP SOAP work against the reference CSMS.

### #256 — accept CS→CP requests without `wsa:From` / `wsa:ReplyTo`

`parseSoapEnvelope` required `From` (and `ReplyTo`) on inbound requests and faulted with `SOAP envelope is missing From`. Apache CXF — and therefore SteVe — omits them on CS→CP calls, so every CSMS→CP operation was rejected. Both headers are only consumed for response routing, and `responseToAddress` already falls back `From → ReplyTo → wsa:anonymous`, so they are now optional. `To`, `MessageID`, and `chargeBoxIdentity` (all of which CXF does send) stay required.

The OCPP-S spec does mark `From` "Required: Yes", but treating a legal, widely-deployed CSMS as unreachable defeats the simulator's purpose; leniency here matches how a real charge point behaves.

### #257 — dispatch the full CS→CP surface for 1.5, and DataTransfer

- **1.5 was 1/15.** `OCPPSoapServer.handleRequest` gated v16-registry dispatch on `isV16Supported || isV12Supported`, so a 1.5 CP fell through to the legacy Reset-only registry. Added a 1.5 version flag; since 1.5 shares 1.6's enum vocabulary it reaches the shared handler registry directly, with no 1.2-style response transform. 1.5 now answers the 14 `target:"cp"` operations in the CP namespace (`urn://Ocpp/Cp/2012/06/`).
- **DataTransfer** (missing on 1.6S and 1.5) is declared with `target:"cs"` + `bidirectional:true`, so it failed the `target === "cp"` dispatch gate even though its handler exists. Dispatch now also admits `bidirectional` operations, and the response serializes in the CP-service namespace. `parseSoapEnvelope` accepts a bidirectional **response** in either service namespace, mirroring the allowance already present for bidirectional requests.

1.2 is unchanged (no DataTransfer in 1.2; its enum-narrowing transform still applies).

## Tests

Because these bugs only surface against a real SOAP CSMS (CXF/SteVe) and not the JSON e2e or hand-written unit envelopes, this adds a **systematic CS→CP dispatch corpus** (`httpServer.soap.bun.test.ts`), data-driven from each dialect's own `operationMetadata`: every operation a dialect declares as CS→CP (`target:"cp"` or bidirectional) must be dispatched, not answered with the not-implemented Fault. It covers the whole 1.2 / 1.5 / 1.6S surface (9 / 15 / 19 ops), so a future dispatch-gate regression is caught across every operation — not just the ones probed by hand. Verified to fail on **18** cases against the pre-fix dispatch, and green after.

Plus focused response-shape tests: 1.5 `GetConfiguration` (asserts the 1.5 CP namespace) and `ChangeAvailability`; 1.6S `DataTransfer`; a `Reset` and a 1.6S `GetConfiguration` each with **no** `From`/`ReplyTo` header (proving #256 leniency reaches both the legacy and the v16 dispatch paths).

Full suites green: `bun test bun.test` (361 pass), full vitest (2117 pass), including the existing SOAP unit + CP→CS client tests that share the modified parser.

## Not included

#178 (a validation warning when `soapCallbackUrl` doesn't end in `/ChargePointService`) ships separately in #259.

### DIFF
--- a/src/cli/server/__tests__/httpServer.soap.bun.test.ts
+++ b/src/cli/server/__tests__/httpServer.soap.bun.test.ts
@@ -14,6 +14,7 @@ import type {
   ParsedSoapEnvelope,
   SoapDialect,
   SoapOperation,
+  SoapOperationMetadata,
   SoapPayload,
 } from "../../../cp/infrastructure/transport/soap";
 import { CPRegistry } from "../CPRegistry";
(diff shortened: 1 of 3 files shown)


shiv3:
@coderabbitai review

PR #264, opened 2026-08-03
fix(security): decode a hex AuthorizationKey before the HTTP Basic header (#260)

### LINKED ISSUE #260: Basic Auth: a hex AuthorizationKey is sent verbatim instead of decoded (OCTT TC_085_CS)
> **Corrected 2026-07-30.** My original report claimed the AuthorizationKey should travel
> verbatim in the Basic header and that pairing with SteVe made a conformant key impossible.
> That was wrong. The OCTT test case document settles it the other way, and the actual defect
> is narrower and clearer. Rewritten below; the original framing is in the comment history.

## Summary

When security profile 1 or 2 is used with a **hex-encoded** `AuthorizationKey`, the simulator
puts the hex string itself in the HTTP Basic header instead of the value it represents. The
OCTT expects the decoded form.

`src/cp/infrastructure/transport/wsUrlWithBasic.ts:79-83`:

```ts
export function buildOcppBasicAuthorization(basicAuth: BasicAuthSettings): string {
  return `Basic ${btoa(`${basicAuth.username}:${basicAuth.password}`)}`;
}
```

`basicAuth.password` is the configured `authorizationKey`, used raw. There is no hex decode
anywhere on this path.

## What the conformance tool requires

*Test case document OCTT for OCPP 1.6* (2025-11), **TC_085_CS — "Basic Authentication - Valid
username/password combination"**, Tool validations, Step 1:

> The authorization header of the HTTP upgrade request must be formatted as follows:
> `AUTHORIZATION: Basic <Base64 encoded(<ChargePointId>:<Configured (non hex representation of) BasicAuthPassword>)>`
>
> Note: The BasicAuthPassword can be configured in two ways:
> 1. When the configured value for BasicAuthPassword is >= 32 and <= 40 characters, the tool
>    will expect that this is the **hex encoded representation** of the password.
> 2. When the configured value for BasicAuthPassword is >= 16 and <= 20 characters, the tool
>    will expect that this is **plaintext (UTF-8)** representation of the password.

So the header carries the **non-hex** value in both cases. A 32–40 character configured value
is hex and must be decoded to 16–20 bytes before it goes on the wire.

This matches the reading the SteVe maintainers converged on in
[steve-community/steve#1895](https://github.com/steve-community/steve/issues/1895): hex when
the key travels over OCPP (`ChangeConfiguration` / `GetConfiguration`), plain when it is used
for HTTP Basic.

## Why the simulator hits this by default

`--authorization-key` **requires** hex (`src/cli/main.ts:401-408`):

```
Error: --authorization-key must be a non-empty hex string
```

So the CLI steers you into exactly the configuration that then goes out un-decoded. Set a
40-hex-digit key — the maximum-entropy form the whitepaper recommends — and the simulator
sends 40 characters where the tool, and a conformant CSMS, expect 20 bytes. Authentication
fails, or silently succeeds against a CSMS that made the same assumption.

Only the 16–20 character plaintext case works today, and it works by coincidence: the raw
string is passed through, which is correct for that mode.

## Two smaller inconsistencies on the same field

**The hex rule is CLI-only.** `cp.create` / `cp.update` (RPC, MCP, web console) go through
`CPRegistry.prepareInit` (`src/cli/server/CPRegistry.ts:333-340`), which checks only that the
key is *present* for profiles 1 and 2. Verified on `0.7.5`:

```jsonc
{"method":"cp.create","params":{"cpId":"ZZ_NONHEX","ocppVersion":"OCPP-1.6J",
 "securityProfile":1,"authorizationKey":"correct horse battery staple!", ...}}
// → accepted, persisted verbatim
```

The same value via `--authorization-key` exits 1. Whichever behaviour is intended, the two
entry points should agree.

**No minimum length, on either path.** The OCTT expects ≥ 16 characters plaintext, or ≥ 32
hex digits. A 2-character key is accepted silently by both. Since the simulator is used to
exercise CSMS implementations, accepting an under-strength key without a word means a CSMS
that *should* reject it looks correct.

## Suggested fix

1. Decode before building the header: if the configured `authorizationKey` is all-hex and of
   even length in the 32–40 range, use its decoded bytes as the Basic password; otherwise use
   it as-is. That covers both OCTT modes.
2. Move the format check into `prepareInit` so the CLI and the RPC behave identically, and
   accept both forms rather than hex only.
3. Warn when the resolved password is shorter than 16 bytes.

Happy to send a PR if the approach looks right to you.

## Environment

- `ghcr.io/shiv3/ocpp-cp-simulator:0.7.5`, daemon mode with `--state-db`
- OCPP-1.6J, security profile 1
- SteVe 3.14.0 as the reference CSMS

shiv3:
## Summary

Fixes #260 (thanks for the corrected analysis and the TC_085_CS citation). With security profile 1/2, the simulator put the configured `AuthorizationKey` **verbatim** into the HTTP Basic header. Per the OCPP 1.6 Security Whitepaper / OCTT **TC_085_CS**, the value on the wire is the **non-hex** form:

- **32–40 chars** ⇒ hex representation of a 16–20 byte password → decode to those bytes before base64.
- **16–20 chars** ⇒ already plaintext UTF-8 → use as-is.

Because the CLI's `--authorization-key` requires hex, the recommended max-entropy 40-hex-digit key went out un-decoded (40 chars where the tool and a conformant CSMS expect 20 bytes), so authentication failed — or silently succeeded against a CSMS that made the same mistake. This matches the reading in steve-community/steve#1895: hex when the key travels over OCPP, plain when used for HTTP Basic.

## The fix

The decode happens once, in `resolveBasicAuth`, which is the single point that feeds **both** the Node `Authorization` header (`buildOcppBasicAuthorization`) and the browser `ocpp_ws_secret` query param (`buildOcppWebSocketUrl`) — so both stay consistent. A new exported helper `authorizationKeyToBasicPassword` applies the length rule; a value outside the 32–40 hex range (or odd-length, or non-hex) passes through unchanged, so short test/legacy keys are unaffected.

Deliberately **untouched**: the `AuthorizationKey` copy served over OCPP `GetConfiguration`/`ChangeConfiguration` keeps its hex form. The two representations were already cleanly separated in `ChargePoint` (the config store is never read back to build the header), so decoding at the header-derivation site doesn't affect the config value.

## Tests

`wsUrlWithBasic.test.ts`: unit coverage for `authorizationKeyToBasicPassword` (32→16 bytes, 40→20 bytes, plaintext passthrough, short/odd/out-of-range passthrough) plus an end-to-end `buildOcppWebSocketConnectOptions` case asserting the decoded `Authorization` header for a 40-hex key. Verified the integration test fails against the pre-fix (verbatim) code. Existing profile-1/2 tests use short keys and stay green. Full `bun test bun.test` 414 pass; `tsc -b` clean on the touched file.

Note the CLI still accepts any hex length; adding a 16–20 / 32–40 length validation (with a clear error) would be a reasonable follow-up but isn't required for the decode fix and would churn existing short-key fixtures.

### DIFF
--- a/src/cp/infrastructure/transport/wsUrlWithBasic.test.ts
+++ b/src/cp/infrastructure/transport/wsUrlWithBasic.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, it } from "vitest";
 
 import {
+  authorizationKeyToBasicPassword,
   buildOcppBasicAuthorization,
   buildOcppWebSocketConnectOptions,
   buildOcppWebSocketUrl,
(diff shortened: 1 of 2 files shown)


shiv3:
@coderabbitai review

Issue #281, opened 2026-09-03
v0.7.6 image never published (release build failed) — could v0.7.7 be released?


juherr:
## Summary

The **v0.7.6 container image was never published**: the release build failed, so GHCR still stops at `0.7.5`. Since **v0.7.7** is already drafted and is `0.7.6` plus one commit, publishing it would deliver both at once — and its commit is the one I am waiting for.

## What happened

The `Build & publish Docker image` run triggered by the `v0.7.6` tag failed:

- https://github.com/shiv3/ocpp-cp-simulator/actions/runs/33736374976 (2026-09-03T08:59:41Z, `failure`)

```
#29 11.30 error: Fail extracting tarball for "@swc/core-linux-arm64-gnu"
#29 15.23 error: Fail extracting tarball from @swc/core-linux-arm64-gnu
#29 ERROR: process "/bin/sh -c bun install --frozen-lockfile --ignore-scripts" did not complete successfully: exit code: 1
ERROR: failed to build: failed to solve: ...
```

It looks transient — the same `bun install` succeeded on `main` eight minutes earlier (08:53) and again at 09:22 — so a re-run of the failed job would most likely be enough.

Current state of the registry (anonymous pull token, `/v2/.../tags/list`):

```
0.7.0, 0.7, 0.7.2, 0.7.3, 0.7.4, 0.7.5
```

No `0.7.6`, and the moving `0.7` tag is still on `0.7.5`. `0.7.1` is missing too, which suggests this failure mode has bitten before.

## The ask

Two options, and the second is the one I care about:

1. Re-run the failed job on the `v0.7.6` tag — that republishes 0.7.6.
2. **Publish v0.7.7.** Its draft contains exactly one commit on top of 0.7.6: `fix(security): security profile 1 keeps the configured ws:// or wss:// scheme (#277)`. That fix is what I need — I deploy this simulator behind a TLS-terminating reverse proxy, and profile 1 rewriting a configured `wss://` into `ws://` is the behaviour that blocked the public path for me.

Option 2 supersedes option 1 for my use case, but of course 0.7.6 being reachable on its own tag has value for anyone pinning it.

## Why it matters on my side

[...]

PR #291, opened 2026-09-04
ci(docker): verify the pushed tags resolve after a tag build (#287)

### LINKED ISSUE #287: Verify the image exists after a tag build, so release notes cannot promise one that is missing
## Summary

#282 makes `bun install` retry, which removes the cause of #281. What it does not do is **verify the outcome**: `.github/workflows/docker-publish.yml` ends at *Build & push*, so a release whose build fails for any other reason still publishes notes that say

> Container image: `ghcr.io/shiv3/ocpp-cp-simulator:X.Y.Z` (published when this release is tagged).

with nothing behind the sentence. The gap is invisible from GitHub — the release page looks complete, the tag exists, the notes promise an image — and shows up only when someone queries the registry. It cost `0.7.6` a same-day re-run, and `0.7.1` was never republished at all.

## Suggestion

After *Build & push*, on tag builds, ask the registry for the tags the release just promised and fail the job if they are not there:

```yaml
- name: Verify the pushed tags resolve
  if: startsWith(github.ref, 'refs/tags/')
  run: |
    for tag in ${{ steps.meta.outputs.tags }}; do
      docker buildx imagetools inspect "$tag" > /dev/null \
        || { echo "::error::$tag did not publish"; exit 1; }
    done
```

A few seconds, no extra credentials, and a silent gap becomes a red release job. An anonymous `GET /v2/<repo>/manifests/<tag>` against `ghcr.io` does the same if you would rather not depend on buildx being present at that point.

A cheaper variant, if you would rather not add a step at all: make the release-notes line conditional on the publish job succeeding, so the notes never claim more than happened.

## Why bother, given the retry

The retry addresses one cause — a flaky tarball. The class is larger: a genuine build failure, a registry outage, a permissions change, an action bump that silently pushes nothing. In all of those the release notes still promise an image. The verification step is what makes "the release says it, therefore it is there" actually true.

Follow-up to #281, where you fixed the immediate cause the same day.

shiv3:
Closes #287.

`Build & push` succeeding is not the same fact as the tags being in the registry, and the release notes promise the registry. The workflow now ends a **tag build** by asking for every tag it just pushed:

```yaml
- name: Verify the pushed tags resolve
  if: startsWith(github.ref, 'refs/tags/')
  run: |
    set -euo pipefail
    missing=0
    while IFS= read -r tag; do
      [ -n "$tag" ] || continue
      if docker buildx imagetools inspect "$tag" >/dev/null 2>&1; then
        echo "ok       $tag"
      else
        echo "::error::$tag was not published -- the release notes promise it"
        missing=$((missing + 1))
      fi
    done <<< "${{ steps.meta.outputs.tags }}"
    if [ "$missing" -ne 0 ]; then
      echo "$missing tag(s) missing from the registry." >&2
      exit 1
    fi
```

Your `buildx imagetools` form, kept as written — `docker/setup-buildx-action` already runs earlier in the job, so it is available at that point, and it needs no extra credentials. Tag builds only: a branch build promises nobody anything, and `latest` moving is not a claim in a release note.

Taking the "why bother, given the retry" point directly: #282 removed one cause. This covers the class — a genuine build failure, a registry outage, a permissions change, an action bump that silently pushes nothing. In all of those the notes still promise an image.

I did not take the cheaper variant (making the notes line conditional). The notes are generated by Release Drafter from a template that runs before the image build, so making the sentence conditional means teaching the release-notes step about another job's outcome — more moving parts than one verification step, and it would still leave the registry unchecked.

## Verification

No buildx locally, so the step's shell was exercised against a stubbed `docker`:

| input | result |
|---|---|
| every tag resolves | prints `ok` per tag, exits 0 |
| one tag missing | prints `::error::` naming that tag, exits 1 |
| trailing blank line in the tag list | skipped, not probed |

`while … done <<< "$TAGS"` is a herestring rather than a pipe, so `missing` survives the loop in the job's shell — worth stating, because the pipe form would have counted correctly and exited 0.

I also confirmed the registry answers the way the step assumes, anonymously: `0.7.8` → `HTTP 200`, `0.7.1` → `HTTP 404`. So had this step existed, `0.7.1` and `0.7.6` would both have been red release jobs.

The wiki's [Image tags](docs/entities/docker-image.md#image-tags) section records the guarantee, with a #281/#287 row in [`docs/sources/github-issues.md`](docs/sources/github-issues.md) and a [`docs/log.md`](docs/log.md) entry.

The first real exercise of the step is the next `vX.Y.Z` tag; I will watch that run.

https://claude.ai/code/session_01FaGcGSwBtuTVJuGWbWFjMw


shiv3:
@coderabbitai review

shiv3:
@coderabbitai review

Issue #295, opened 2026-09-04
`cp.create` creates one charge point at a time — no way to stand up a fleet


shiv3:
## Summary

`cp.create` creates exactly one charge point from an explicit `cpId`. There is no way to say "give me 50 charge points against this CSMS", so every fleet-shaped use has to loop the RPC from the outside, invent its own id scheme, and repeat the whole parameter block per call.

That block is not small — `cpParamsBaseSchema` in `src/protocol/methods.ts` carries `wsUrl`, `ocppVersion`, `connectors`, `vendor`/`model`, the four SOAP fields, `securityProfile` / `authorizationKey`, the TLS material and `bootNotification`. Repeating it 50 times over socket.io, or building it 50 times in an MCP client, is exactly the friction #111 is trying to remove: a CSMS project that wants to check its own behaviour under more than one station currently writes the fan-out loop itself.

## Proposal

A separate method rather than an overload of `cp.create` — the single-CP call returns one CP object and callers depend on that shape.

```
cp.create_many {
  count: number,          // 1..limit
  idPattern: string,      // "CP{n:03}" -> CP001, CP002, ...
  startIndex?: number,    // default 1
  ...cpParamsBase         // every existing cp.create field, shared by all
}
-> { created: string[], failed: { cpId: string, reason: string }[] }
```

Two behaviours worth pinning down now, because they are the ones that bite later:

- **Partial success is the result, not an error.** One CP failing to connect should not roll back 49 good ones. Hence the `created` / `failed` split rather than a throw.
- **Creation is sequential**, so registry `event` pushes arrive in id order and a subscriber can rely on that.

## Notes

- Reuse `cpParamsBaseSchema` directly. Restating a subset is what made `cp_create` narrower than `cp.create` in #284; deriving the MCP tool from the same schema is the fix that should hold here from the start.
- The `count` ceiling belongs in `src/protocol/limits.ts` and should be enforced, not just documented — this is the one method that can allocate unbounded resources in a single call.
- CLI equivalent: `--cp-count` / `--cp-id-pattern` alongside the existing `--cp-id`.
- Dispatch goes in `dispatchRpcCore` (`src/cli/server/socketServer.ts`) and the loop in `src/cli/server/CPRegistry.ts`.

## Acceptance

- `cp.create_many` with a bad `wsUrl` on one generated id still creates the rest and reports the failure.
- The MCP tool is derived from the zod schema, not hand-written.
[...]

PR #305, opened 2026-09-04
feat(transport,control-plane): let a charge point hold several supervision URLs (#296)

### LINKED ISSUE #296: A charge point is pinned to one supervision URL, so a dead CSMS node blocks it forever
## Summary

A charge point is pinned to a single `wsUrl` for its whole life. If the CSMS under test sits behind more than one endpoint — a load balancer with several nodes, a blue/green pair, a cluster where the point of the test is that the CP survives one node going away — there is no way to express that. The CP reconnects to the same URL forever, so a dead node blocks that CP permanently instead of draining to a healthy one.

`src/cp/infrastructure/transport/OCPPWebSocket.ts` already does the hard part: it has exponential backoff with a reconnect reservation. What it does not have is anywhere else to go.

## Proposal

`wsUrl` accepts `string | string[]`, plus a distribution policy:

```
wsUrl: string | string[]
urlDistribution?: "round-robin" | "random" | "cp-affinity"   // default "round-robin"
```

- `round-robin` — rotate on each reconnect attempt.
- `random` — draw per attempt.
- `cp-affinity` — hash the `cpId` to a fixed URL, so a CP lands on the same node across restarts. This is the one policy that keeps a reconnect deterministic, which matters if a test asserts on which node saw the session.

Selection happens at connect time, so a reconnect after a failure moves to the next URL rather than retrying a node that is gone.

## Notes

- Single-string `wsUrl` keeps working unchanged; this is purely additive to `cpParamsBaseSchema` in `src/protocol/methods.ts`.
- Worth deciding explicitly: does backoff reset when the URL changes, or keep climbing? I'd keep climbing — otherwise a fleet pointed at a fully dead cluster hot-loops through the URL list at the base delay.
- Pairs naturally with fleet creation, where "spread N CPs across 3 CSMS nodes" is the actual use case.

## Acceptance

- Unit test: after a forced disconnect (reuse the network simulation fault injection in `src/cp/infrastructure/transport/network-sim/`), the CP reconnects to a different URL under `round-robin` and to the same URL under `cp-affinity`.
- The chosen URL appears in the connection log line so a failure is diagnosable.
- `docs/concepts/control-plane.md` and `docs/concepts/ocpp-versions-and-transports.md` updated in the same PR.

shiv3:
Closes #296.

## Summary

A charge point was pinned to one `wsUrl` for its whole life. A CSMS behind more than one endpoint — a load balancer, a blue/green pair, a cluster where the *point* of the test is that the CP survives losing a node — could not be expressed, and the reconnect loop retried the same dead URL forever.

`wsUrl` now accepts a list, with `urlDistribution` choosing between them:

```jsonc
{"method":"cp.create","params":{
  "cpId":"CP001",
  "wsUrl":["wss://node-a/ocpp/","wss://node-b/ocpp/","wss://node-c/ocpp/"],
  "urlDistribution":"cp-affinity"}}
```

`OCPPWebSocket` already had the hard part (exponential backoff with a reconnect reservation); what it lacked was anywhere else to go.

## Scoped so nothing downstream changes

The list stays in the charge point's config and **one URL is resolved per connection attempt**. `cp.list`, the persisted `charge_points.ws_url` and every log line keep seeing a single string — the one currently in play. No DB schema change, no restore-path change.

A list is **OCPP-J only**: SOAP posts to the Central System service and is called back on one advertised address, with no reconnect loop to rotate, so it is refused rather than accepted and ignored.

## `cp-affinity` is sticky, deliberately

The two obvious readings contradict each other once the primary is down: always return the primary (and never connect), or rotate like round-robin (and lose the determinism that is the point). So the primary is retried while consecutive failures are below the threshold (3), the pool advances one URL on reaching it, and **any success resets it to the primary** — the next disconnect episode starts from the assigned node again, so the pool never has to probe a node it is not talking to. Only a close the CP did not ask for counts as a failure; a manual `disconnect` says nothing about the node.

The policy lives in `SupervisionUrlPool` — pure and synchronous, so the failover semantics are testable without a socket.

## Review

`codex review --base main` raised two findings; both were real and are fixed here:

- **P1 — the feature was inert over the control plane.** `parseCreateBody` produced the list and `RegistryChargePointService.toInitOptions()` dropped it, so the charge point came up on the first URL with no pool while `cp.create` answered success. My unit tests passed throughout because they only exercised `parseCreateBody` in isolation; there is now a test that drives create *and* update through the facade and asserts what actually reached the charge point.
- **P2 — `current()` consumed the RNG under `random`**, and it is called at wiring time, so the first seeded draw was discarded and merely inspecting the pool changed which node the next attempt used. The selected index is held now, and a test pins that inspection is side-effect free.

## Checks

`bun run lint`, `bun run format:check`, `bun build --compile` smoke, `bun run test:vitest` (2279 pass), `bun run test:bun` (447 pass), wiki link lint (0 problems). 23 new tests across `SupervisionUrlPool.test.ts` and `supervisionUrls.test.ts`.

Docs in the same commit per `CLAUDE.md`: `control-plane.md` (params rows plus a "Multiple supervision URLs" section covering the resolve-per-attempt scoping, the affinity contract and the SOAP refusal), `log.md`, `github-issues.md`.

> Independent of #295 and branched from `main`, but both touch `src/protocol/methods.ts` in different regions — merging either order is fine.

https://claude.ai/code/session_014udBrUFZpA5WqirTnCzMGH

## Summary by CodeRabbit

- **New Features**
  - OCPP-J charge points can use multiple supervision URLs.
  - Choose round-robin, random, or charge-point-affinity URL selection with automatic failover.
  - URL settings persist across service restarts.
  - Added batch charge-point creation with configurable ID patterns and partial-success handling.
  - SOAP charge points continue to support only a single supervision URL.

- **Documentation**
  - Added guidance for URL lists, distribution policies, failover behavior, persistence, and batch creation.


shiv3:
## Follow-up: three ways the URL list was present in the API and absent where it mattered

CodeRabbit flagged the restart semantics as a documentation gap. Chasing it found a **real bug**, and re-running `codex review` on the fix found two more. All three are fixed in 09840da.

**1. The list did not survive a restart (schema v7).** `charge_points` stored only `ws_url` and `restoreFromDatabase` restored only that, so a charge point created with a list came back from `--state-db` with failover silently disabled — the one thing the list exists to provide, gone, with nothing to say so. Added `supervision_urls` (JSON) and `url_distribution` with a v6→v7 migration. `ws_url` still holds the single URL every other reader expects, so no existing reader changes.

**2. An update could not clear the pool.** `toInitOptions` fell back to the stored list when an update omitted one — but `wsUrl` is *required* on `cp.update`, so an update fully specifies the URL configuration. A charge point re-pointed at a single endpoint kept failing over to the stale pool, and kept persisting it.

**3. The typed remote adapter dropped the list.** `CreateChargePointParams` carries `supervisionUrls` in a field of its own because that is what the charge point consumes, but the wire expresses it as an array `wsUrl`. `RemoteChargePointService` forwarded the typed shape verbatim, so the schema stripped the unknown key: the call succeeded and quietly created a single-URL charge point.

Six new tests cover the restart round-trip (including the single-URL case, which must stay untouched), the clear-on-update semantics, and the adapter translation.

`bun run lint`, `format:check`, compile smoke, `test:vitest` (2268), `test:bun` (465). `codex review --base main` on the final state reports no blocking correctness issues.

Issue #320, opened 2026-09-04
The published CLI tarball cannot start: package.json#files omits src/data, src/ocpp, src/trace and vendor/


shiv3:
## Summary

The CLI tarball `cli-release.yml` publishes cannot run. It dies on the **first import** of `main.ts`, so `--version` and `--help` both fail:

```
$ ocpp-cp-sim --version
error: Cannot find module '../data/sqlite/SqliteConnectorSettingsRepository'
  from '/private/tmp/rel-bun-g/node_modules/ocpp-cp-simulator/src/cli/main.ts'
VERSION_EXIT=1
```

Installed exactly the way `cli-release.yml` advertises (`bun install -g <tgz>` into a sandboxed global dir). The install itself exits 0 — the failure is at runtime.

## Cause

`package.json#files` is:

```json
["src/cli", "src/cp", "src/protocol", "src/scenario",
 "src/utils/scenarioTemplates.ts", "src/utils/scenarios", "schema", "dist"]
```

It omits **`src/data`, `src/ocpp`, `src/trace` and `vendor/`** — all four of which `Dockerfile` copies with comments explicitly stating the daemon crashes without them. `src/cli/main.ts:37` imports `../data/sqlite/SqliteConnectorSettingsRepository`.

```
$ npm pack --silent --ignore-scripts     # exit 0 → ocpp-cp-simulator-0.0.0.tgz, 1.8M
$ tar -tf ocpp-cp-simulator-0.0.0.tgz | grep -cE 'package/(src/(ocpp|data|trace)/|vendor/)'
0
```

## Why CI does not catch it

`cli-release.yml`'s three verify steps only `grep` the tarball listing for `src/protocol/`, `src/scenario/scenarioSchemaValidator.ts` and `schema/scenario.schema.json` — all present, all passing. **Nothing ever boots the package.** And `grep -n "pack\|tgz\|install -g" .github/workflows/ci.yml` returns nothing: no CI job installs the tarball at all.

So the guards are structurally incapable of seeing this class of defect: they check that named files are present, not that the thing runs.

## Since when, and why nobody noticed

Broken at least since `8a0fda5` (2026-07-01, "route standalone CLI through the facade"), which introduced the `main.ts → ../data/` import. The last CLI release, **`cli-v0.3.1`, is from 2026-06-02** and predates it — installing that pinned tarball works fine (`--help` exit 0). So the published artifact has never been broken; the *next* one would be.

## Acceptance

[...]

PR #324, opened 2026-09-04
fix(desktop): the bundled daemon could never find the web console (#319)

### LINKED ISSUE #319: Desktop app: the bundled daemon exits(1) on spawn — broken since v0.3.2, including shipped v0.7.8
## Summary

Every desktop release since **v0.3.2** ships a daemon that exits(1) immediately on spawn. The user sees the splash screen poll for 30 s and then render **"Daemon failed to start"**. That is the entire product.

This was verified against the **published v0.7.8 macOS bundle**, not just the source:

```
$ "OCPP CP Simulator.app/Contents/MacOS/ocpp-cp-simulator"
[daemon] Error: --web-console requires the bundled UI to be built. Run `bun run build` in the repo first (or use an installed package which ships dist/).
[daemon] exited: TerminatedPayload { code: Some(1), signal: None }
```

Reproduced identically against the shipped **v0.7.5** bundle, and against a sidecar built from current `main` with `scripts/build-tauri-sidecar.sh`.

## Cause

`src-tauri/src/lib.rs:74` spawns the sidecar with `--web-console`. `resolveBundledDist()` (`src/cli/main.ts:52`) is:

```ts
const candidate = path.resolve(import.meta.dir, "../../dist");
```

Inside a `bun build --compile` binary, `import.meta.dir` is `/$bunfs/root`, so the candidate resolves to `/dist` — which never exists. Confirmed directly:

```
import.meta.dir=  /$bunfs/root
import.meta.path= /$bunfs/root/t
execPath=         /private/tmp/rel-fs/t
```

The lookup needs to be relative to `process.execPath` (or the bundled asset needs to be embedded), not to `import.meta.dir`.

## Isolation

The same shipped binary is healthy without that one flag, which also shows `release.yml`'s version stamping works correctly:

```
$ .../Contents/MacOS/ocpp-cp-sim --http-host 127.0.0.1 --http-port 19707 --state-db ... --log-format json
{"ok":true,"version":"0.7.8"}   HTTP=200
```

## Scope

`git tag --contains` for both `38d3b2a` (added `--web-console` to `lib.rs`) and `241cff4` (added `resolveBundledDist`) is **v0.3.2**. So roughly 30 desktop releases are affected. No existing issue tracks it. (v0.7.0 shipping zero assets is a separate, already-known symptom of the old tauri-action draft bug.)

## Acceptance

- A freshly built desktop bundle starts its daemon and serves the web console.
- Something in CI or the release job actually **launches** the bundled app (or at minimum runs the sidecar with the exact args `lib.rs` passes) rather than only building it — this shipped for 30 releases precisely because nothing ever ran it.

shiv3:
Closes #319.

## The bug

Every desktop release since **v0.3.2** — roughly 30 of them, including the
currently published **v0.7.8** — ships a daemon that exits(1) the moment Tauri
spawns it. The splash screen polls `/v1/healthz` for 30 s and then renders
**"Daemon failed to start"**. That is the entire user-visible product.

Confirmed against the *published* v0.7.8 and v0.7.5 macOS bundles, not just
from source:

```
$ "OCPP CP Simulator.app/Contents/MacOS/ocpp-cp-simulator"
[daemon] Error: --web-console requires the bundled UI to be built...
[daemon] exited: TerminatedPayload { code: Some(1), signal: None }
```

The same shipped binary is healthy without that one flag, which also shows the
release version stamping works: `{"ok":true,"version":"0.7.8"}`.

## Root cause — and it was two things, not one

`resolveBundledDist()` resolved `dist/` from `import.meta.dir`. Inside a
`bun build --compile` binary that is `/$bunfs/root`, so the candidate was
`/dist`, which never exists.

But fixing the path alone would not have been enough. Tracing where `dist/`
actually lives in each of the three runtime shapes:

| Shape | Where `dist/` is |
| --- | --- |
| Source checkout | `<repo>/dist` — reachable from `import.meta.dir` |
| Installed package | `<pkg>/dist` — same two levels up |
| **Tauri bundle** | **nowhere on disk** |

`frontendDist` is embedded *in the Rust binary* (the Tauri config schema says a
relative path "is read recursively and all files are embedded in the
application binary"), and `tauri.conf.json` had no `bundle.resources` entry —
so the sidecar had no file to find at any path.

## The fix

- `resolveBundledDist()` searches `import.meta.dir/../../dist` plus three
  `process.execPath`-relative candidates (`<execDir>/dist`, `<execDir>/../dist`,
  `<execDir>/../Resources/web-console`). `process.execPath` is the real path on
  disk, which `import.meta.dir` is not in a compiled binary.
- `tauri.conf.json` gains `"resources": {"../dist/": "web-console/"}`, so the
  file exists to be found.
- `lib.rs` resolves `resource_dir()/web-console` and passes it explicitly; the
  flag pair is dropped entirely when there is no `index.html` there, so
  `tauri dev` still works through the `bun src/cli/main.ts` branch.
- New `--web-console-dist <dir>` flag, authoritative: a wrong explicit path is
  an **error**, never a silent fallback.
- **The error message was part of the bug.** "Run `bun run build` in the repo
  first" is useless to someone who double-clicked a `.dmg`. It now prints every
  path searched, plus `import.meta.dir` and `process.execPath`, and offers a fix
  per runtime shape.

## The test that would have caught it

`src/build/__tests__/tauriSidecarWebConsole.bun.test.ts` compiles the binary
once and then **launches** it. Two properties make it resistant to the way this
survived 30 releases:

- **The arguments are parsed out of `lib.rs`'s `DAEMON_ARGS` template**, not
  copied into the test. A test with its own copy would pass while `lib.rs`
  drifted away from it — the same blindness in a new place.
- **`POLL_TIMEOUT_MS` and the health path are read from `public/splash.html`**,
  so readiness is judged against the user's actual budget. A process that exits
  1 and one that never becomes ready look identical to the user; both fail.

Six cases: the template-parse guard, the resource-dir layout, `dist/` beside the
binary, the macOS `.app` layout, the error naming its searched paths, and a bad
explicit path. ~12 s, under `bun run test:bun`, so it gates every PR.

## Testing

Gate: `lint` ✅, `format:check` ✅, `tsc --noEmit` **exactly 193 — identical to
`main`**, `bun build --compile` ✅, `vite build` ✅, `vitest` 203 files / 2373
passed ✅, `bun test` 530 passed ✅, wiki link lint 0 problems ✅.

Six mutation checks, each restored and re-verified. The one that matters:
reverting `resolveBundledDist` to its `import.meta.dir` one-liner fails 5 of the
6 cases with `searched: /dist` and `import.meta.dir: /$bunfs/root` — the v0.7.8
root cause, reproduced by the test.

## Not verified here

- **`tauri build` itself** — no Rust toolchain on this host. The Rust does not
  compile here; it is hand-checked against `Command::args`, `resource_dir()` and
  `Option<PathBuf>::as_deref()`.
- That `{"../dist/": "web-console/"}` lands at `<resource-dir>/web-console/` on
  all three platforms. If a platform's layout differs, the flag is simply
  dropped and the daemon falls back to its own search plus a diagnostic — a bad
  path can no longer take the app down silently, but the console would not be
  served there.
- The desktop `dist/` is now shipped **twice** (embedded for `splash.html`,
  resource for the daemon). Deduplicating means serving over `tauri://` instead
  of HTTP — a different architecture, out of scope here.

Separately found and **not** fixed here: `package.json#files` is insufficient to
run the CLI from a published tarball. That is #320, fixed in its own PR.

https://claude.ai/code/session_01BZS1CS7fSa5sSgnUfbsoZq


shiv3:
@coderabbitai review

shiv3:
@coderabbitai review

PR #325, opened 2026-09-04
fix(cli,ci): the published tarball could not start, and its install URL 404s (#320, #321)

### LINKED ISSUE #320: The published CLI tarball cannot start: package.json#files omits src/data, src/ocpp, src/trace and vendor/

### LINKED ISSUE #321: The documented releases/latest install URL 404s (two tag trains share "latest")
## Summary

The install command in `README.md` (lines 36, 39), `docs/entities/cli.md:52`, and in the release notes `cli-release.yml:129` generates, 404s right now:

```
$ bun install -g https://github.com/shiv3/ocpp-cp-simulator/releases/latest/download/ocpp-cp-simulator.tgz
error: GET https://github.com/shiv3/ocpp-cp-simulator/releases/download/v0.7.8/ocpp-cp-simulator.tgz - 404
error: ... failed to resolve
```

## Cause

This repository has **two tag trains** — `v*` (desktop app) and `cli-v*` (CLI) — and `releases/latest` resolves across both. The newest release is currently `v0.7.8`, a desktop release, which carries no `.tgz`.

So the documented URL works only while the most recent release of *either* kind happens to be a `cli-v*`. Cutting a CLI release fixes it until the next desktop tag breaks it again. It is not a transient state; it is the steady state for half the time.

`CLAUDE.md` treats documented instructions as contracts, so a quick-start command that does not run is a defect rather than a stale note.

## Options

- Point at a pinned `cli-v*` tag in the docs and update it on each CLI release (verified working: the `cli-v0.1.0` URL installs and `--help` exits 0).
- Or publish the `.tgz` as an asset on the desktop releases too, so `latest` always carries one.
- Or drop `latest` from the documented URL in favour of a redirect the project controls.

Whichever is chosen, `cli-release.yml`'s own generated release-notes body needs the same treatment — it currently advertises the broken form.

## Acceptance

- The command printed in `README.md` and `docs/entities/cli.md` runs successfully at any point in the release cycle, not only just after a CLI release.
- Ideally something checks it, so this is caught by CI rather than by a user.

Related: #320 (the tarball that URL points at cannot currently start).

shiv3:
Closes #320. Closes #321.

Two defects in the same surface — how the CLI is distributed — so one commit.

## #320: the published tarball could not start

`package.json#files` omitted `src/data`, `src/ocpp`, `src/trace`, `src/utils`
(beyond two narrow entries) and `vendor/`. An installed package died on the
first import:

```
$ ocpp-cp-sim --version
error: Cannot find module '../data/sqlite/SqliteConnectorSettingsRepository'
  from '.../node_modules/ocpp-cp-simulator/src/cli/main.ts'
```

Broken since `8a0fda5` (2026-07-01). The last CLI release, `cli-v0.3.1`,
predates it and installs fine — so the published artifact was never broken; the
**next** one would have been.

### The set was established empirically, not copied

A mutation matrix removed each candidate entry one at a time, packed, installed
and ran. **All nine source/data entries fail `--help`** — every one is on
`main.ts`'s eager static-import path — so the final set is minimal *and*
complete. Two things the issue's own list did not have:

- `src/utils/scenarioTemplates.ts` + `src/utils/scenarios` is **not enough**:
  it fails with `Cannot find module '../../utils/blueprints'` from
  `socketServer.ts`. Widened to the whole `src/utils`, matching the Dockerfile.
- `vendor/ocpp-schemas/NOTICE` ships alongside the schemas — verbatim and
  unedited, as CC BY-ND requires.

All eleven entries are directories now. Enumerating individual files is what let
this drift in the first place.

## The verification is the actual fix

`cli-release.yml`'s three verify steps only `grep`ped the tarball listing for
filenames. Every one of them **passed while the package was unusable**, and
`ci.yml` touched the tarball not at all — so this breaks on a PR and is
discovered at release time, six weeks later.

New `scripts/verify-cli-tarball.sh` installs into a throwaway prefix and
**boots** it:

| Stage | What it proves |
| --- | --- |
| A | `--help` exits 0 and prints the usage banner |
| B | `bun build --compile` over the *installed* tree — resolves the whole graph including type-only imports |
| C | daemon with a bootstrapped CP → `/v1/healthz` 200, process still alive |
| D | `--web-console` serves the UI shell (gated on `--expect-dist`) |

The greps are gone; the script replaces them, and `ci.yml` runs it on **every
PR**.

**Mutation results.** Dropping any of the nine source directories fails stage A.
Dropping `dist` passes A, B **and** C — only stage D catches it. That is the
evidence the layering is not decorative. Honest caveat: because every source
path is eagerly imported today, B and C are not the discriminating check for any
*current* mutation; they guard the class (lazy or type-only imports, and
breakage that is runtime rather than import-time).

## #321: the documented install URL 404s

`releases/latest` resolves across **both** tag trains (`v*` desktop and
`cli-v*`), so with `v0.7.8` newest the documented command fails — and would
fail for roughly half of every release cycle.

**Chosen: a rolling `cli-latest` pre-release.** `cli-release.yml` force-moves
the tag and `--clobber`s the asset on each CLI release; README, `cli.md` and the
generated release notes all point at
`releases/download/cli-latest/ocpp-cp-simulator.tgz`.

Why not the alternatives: pinning a `cli-v*` tag in the docs needs a human to
remember, which is exactly why the current form fails; attaching the `.tgz` to
desktop releases couples the two trains and publishes CLI tarballs under
desktop version numbers.

**Marked pre-release deliberately** — GitHub never resolves `releases/latest` to
a pre-release, so the new pointer cannot hijack the desktop badge. Verified
`cli-latest` matches no tag trigger: only `cli-v*` and `v*` exist across all
workflows.

`ci.yml` now asserts the three places carry the same URL, and bans any
`install -g …releases/latest/download` across tracked `*.md` and `.github`.
Mutation-checked both ways.

## Testing

Gate: `lint` ✅, `format:check` ✅, `tsc --noEmit` **exactly 193 — baseline**,
`bun build --compile` ✅, `vite build` ✅, `vitest` 203 files / 2373 ✅,
`bun test` 523 ✅, `verify-cli-tarball.sh` and `--expect-dist` both ✅, wiki
link lint 0 problems ✅.

Proof the tarball now works, run the way the bug was found:

```
$ npm pack --silent --ignore-scripts   → exit 0, 2.1M
$ BUN_INSTALL=… bun install -g …tgz    → exit 0
$ …/bin/ocpp-cp-sim --help             → exit 0, 176-line usage banner
```

## Known window, stated plainly

**The new URL 404s until the first `cli-v*` release after this merges** — the
rolling release does not exist yet. Every option in #321 has that property. The
pinned line in the docs (updated `cli-v0.1.0` → `cli-v0.3.1`) works right now:
probed HTTP 200, while the old `releases/latest` form redirects to `v0.7.8` and
returns 404. Cutting a CLI release closes the window.

Unproven until a real release: `git push -f` of `refs/tags/cli-latest` under
`GITHUB_TOKEN` (fails if a tag ruleset restricts updates), `--clobber` replacing
a same-named asset, and CDN lag after it (the 5×10 s retry is a guess). The
"verify the documented install URL" step is a **post-publish canary** — if it
fails the release is already out, so the blocking check is the local one that
runs before publishing.

Related: #319 / #324 — the same "nothing ever ran the artifact" failure, in the
desktop bundle.

https://claude.ai/code/session_01BZS1CS7fSa5sSgnUfbsoZq

---

## Merge note: what has and has not been exercised

**Seven rounds of `codex review`; the last returned no findings.** Local gate, `verify-cli-tarball.sh`, `verify-install-urls.sh` and GitHub Actions are all green, and both new CI steps have now run on a runner.

**Not exercised against real GitHub.** Every `roll-pointer` path is verified against a stubbed `gh` that models its interface and two of its semantics. These remain unproven until a maintainer cuts a real `cli-v*` release:

- the asset **rename** endpoint's real behaviour, and `delete-asset`'s 404 message;
- **propagation lag** after a rename;
- `git push -f` of `refs/tags/cli-latest` under `GITHUB_TOKEN`, if a tag ruleset restricts updates;
- the release **listing's** actual consistency window.

**The first real CLI release is the acceptance test for those.** The workflow's post-publish URL check is a canary, not a gate — if it fires, the release is already out. The blocking checks run before publishing.

**Merging is still the right call**, because `main` today ships:

- a CLI tarball that **cannot start** (`package.json#files` omits four source trees — #320), and
- an install URL in the README that **404s** (`releases/latest` spans both tag trains — #321),

and a pointer mechanism that would use `--clobber`, which deletes the live asset before uploading its replacement. This PR fixes all three and, where it cannot make a failure impossible, makes it fail toward "the previous release keeps working". The one irreducible 404 case — rename *and* recovery both exhausted, with no atomic asset swap available on GitHub — exits loudly saying how to restore.

**CodeRabbit never reviewed this PR.** Its check reports `pass`, but that is the rate-limited state, not a review: the free OSS quota was exhausted on every attempt. Recorded here so the check status is not mistaken for a signal.


shiv3:
@coderabbitai review

PR #354, opened 2026-09-12
feat(cli): add --soap-tunnel ngrok for SOAP callback exposure (#183)

### LINKED ISSUE #183: Add optional ngrok support for SOAP callback endpoints
## Context

SOAP-based OCPP versions require the simulator to expose a callback endpoint for CSMS → CP messages:

```text
/ocpp/soap/<cpId>/ChargePointService
```

Today, users must provide the complete public callback URL explicitly through:

```bash
--soap-callback-url
```

This works well when the simulator and CSMS share the same local network or Docker network, but it is less convenient when testing against a remotely hosted CSMS.

In that case, the callback endpoint must be publicly reachable, which currently requires users to configure a tunnel manually before starting the simulator.

## Suggestion

Add optional native support for exposing the SOAP callback endpoint through ngrok.

For example:

```bash
ocpp-cp-sim \
  --ocpp-version OCPP-1.6S \
  --ws-url https://csms.example.com/CentralSystemService \
  --soap-tunnel ngrok
```

The simulator would:

1. start the local SOAP callback server;
2. start or connect to ngrok;
3. retrieve the generated public URL;
4. automatically build the full callback URL:

```text
https://<generated-domain>/ocpp/soap/<cpId>/ChargePointService
```

5. use that URL in the SOAP `From` / callback configuration;
6. display the resolved public callback URL in the logs and UI.

## Proposed precedence

The callback URL resolution could follow this order:

1. `--soap-callback-url`
   - use the explicitly provided full URL;

2. `--soap-public-base-url`
   - build the callback URL from a user-provided public base URL;

3. `--soap-tunnel ngrok`
   - create a tunnel and derive the callback URL automatically;

4. no public URL configuration
   - use a local URL only when appropriate, or return a clear validation error.

This would keep the current explicit configuration fully supported while making `--soap-callback-url` optional when another callback URL provider is configured.

## Possible CLI options

```text
--soap-callback-url <url>
--soap-public-base-url <url>
--soap-tunnel <none|ngrok>
--ngrok-auth-token <token>
--ngrok-domain <domain>
```

The auth token should preferably also be readable from the standard ngrok configuration or an environment variable, so it does not need to appear in command history.

## Docker considerations

For Docker usage, the ngrok process could either:

- be started by the simulator itself; or
- run as a separate container, with the simulator connecting to the ngrok agent API.

The second option may be easier to operate in containerized environments and avoids embedding ngrok directly into the simulator image.

## Security considerations

Because the tunnel exposes a CSMS → CP command endpoint publicly, the implementation should:

- print a clear warning when the endpoint is publicly exposed;
- preserve the existing strict charge-point identity checks;
- bind the tunnel only to the SOAP callback server;
- stop and clean up the tunnel when the simulator exits;
- avoid logging authentication tokens;
- document that temporary ngrok URLs may change between runs.

## Why this would be useful

This would significantly simplify testing SOAP-based OCPP versions against:

- remotely hosted SteVe instances;
- staging or integration CSMS environments;
- CI or ephemeral test environments;
- CSMS deployments that cannot directly reach the developer's machine.

It would also make the newly supported OCPP 1.2, 1.5 and 1.6S flows easier to try without requiring users to understand and configure reverse tunnels manually.

## Alternative design

To avoid coupling the SOAP implementation specifically to ngrok, the feature could be implemented behind a generic callback exposure abstraction.

Ngrok would then be the first supported provider, while leaving room for other mechanisms later, such as Cloudflare Tunnel or an externally managed tunnel.

juherr:
Fixes #183

## What

`--soap-tunnel ngrok` exposes the daemon's SOAP `ChargePointService` callback endpoint through an ngrok tunnel and uses the tunnel's public origin as the daemon's **SOAP public base**. Every SOAP charge point without a callback URL — bootstrapped from the flags, restored from `--state-db`, or created later over `cp.create` / the web console — gets `{base}{soapPath}/{cpId}/ChargePointService` derived from it. The console shows the resulting URL with a **Copy** button: that is the value to register in the CSMS.

## How

**Bound to the SOAP callback server only, in the issue's order.** The daemon binds a dedicated listener that serves the callback route and answers 404 to everything else (`createSoapCallbackHandlers`; `--soap-tunnel-port`, a free port by default), opens the tunnel towards *that* listener, and only then builds the registry, restores the fleet and binds the API / console listeners. So the tunnel publishes the CSMS→CP endpoint alone — socket.io, the web console, MCP, health and metrics stay local — and a restored charge point never announces a callback URL nothing is bound behind. `startServer` returns a `RunningServer` handle so the ordering is tested against the real daemon.

**No npm dependency.** Two modes behind one `SoapTunnel` seam (`src/cli/soapTunnel.ts`):

- **spawn** (default) — runs the `ngrok` binary from `PATH` (`ngrok http <host>:<port> --log stdout --log-format json`), reads the public URL from its `started tunnel` log line, kills it on shutdown and on process exit (a startup failure past the tunnel leaves nothing behind). If the agent dies on its own the daemon stops with exit code 1: a SOAP charge point without a reachable callback cannot do its job.
- **attach** (`--ngrok-api-url <url>`) — reads `GET /api/tunnels` of an agent someone else runs, typically an `ngrok/ngrok` sidecar next to the Docker image (which ships no ngrok binary). Never stops that agent. Requires `--soap-tunnel-port`, and the agent's https tunnel must forward to exactly that port — a tunnel to any other port (say, the API port) is refused even when it is the only one, naming the port it does forward to.

**Security.** `--ngrok-auth-token` reaches the agent through `NGROK_AUTHTOKEN` in its environment — never on argv, never in a log line — and its value is scrubbed from anything ngrok reports. The startup lines print the tunnel origin, the derived callback URL, a public-exposure warning and the fact that only the callback route is published. Charge-point identity checks on the route are unchanged. The tunnel listener has no Basic Auth gate — the CSMS is its only client and OCPP-S callbacks carry no credentials; the per-charge-point `--basic-auth-*` credentials cover the outgoing connection only.

**Derivation lives in the registry**, at instantiation, so one path covers bootstrap, restore and RPC. A derived URL is flagged `soapCallbackUrlDerived` and **never persisted**: a free-tier tunnel URL changes between runs, so a restored charge point re-derives from the current base. `cp.update` fully specifies the callback URL, like the supervision URLs: an update without one derives, so clearing an explicit URL in the form switches the charge point back to derivation. This also closes a pre-existing gap: `--soap-public-base-url` was not applied to RPC-created charge points either.

**New `server.info` RPC** — `{ version, soap: { publicBaseUrl, path, tunnel } }`. `cp.create` accepts a SOAP charge point without `soapCallbackUrl` when the daemon reports a base.

**Web console (remote mode)** — the callback URL is optional in the create / edit form when the daemon has a base (the derivation is previewed as the placeholder and never sent back as explicit), and the charge point's Configuration tab shows the effective URL with a Copy button, noting when it came from the tunnel.

## CLI

```
--soap-tunnel <none|ngrok>     default none; server mode + SOAP version only
--ngrok-auth-token <token>     spawn only; via env, never logged
--ngrok-domain <domain>        spawn only; `ngrok http --url`
--ngrok-api-url <url>          attach mode; incompatible with the two above
--soap-tunnel-port <port>      local port of the callback-only listener (default 0 = free port; required in attach mode)
```

`--soap-tunnel` is **refused** together with `--soap-callback-url` / `--soap-public-base-url` rather than ranked below them — two sources for one public base is a contradiction, not a precedence.

```bash
ocpp-cp-sim --ocpp-version OCPP-1.6S \
  --ws-url https://csms.example.com/steve/services/CentralSystemService \
  --cp-id CP1 --daemon --soap-tunnel ngrok
# [server] SOAP callback listener on http://127.0.0.1:53412 (tunnel target; callback route only)
# [server] SOAP tunnel (ngrok, spawn): https://a1b2.ngrok-free.app -> http://127.0.0.1:53412
# [server] Warning: SOAP callback endpoint is publicly reachable through the tunnel at https://a1b2.ngrok-free.app/ocpp/soap/CP1/ChargePointService; …
```

## Docs

`docs/concepts/ocpp-versions-and-transports.md` (new *Exposing the callback through a tunnel* section, precedence rewritten, compose sidecar recipe), `docs/entities/cli.md`, `docs/entities/daemon.md`, `docs/entities/web-console.md`, `docs/concepts/control-plane.md` (`server.info`), `docs/sources/github-issues.md`, `docs/log.md`.

## Tests

- `soapTunnel.test.ts` (21, runs under vitest and `bun test`): argv/env, JSON log parsing, error line with redaction, early exit quoting stderr, missing binary, timeout, close/onExit, stdout kept open after start (leaving the loop cancels the pipe and SIGPIPEs ngrok — caught end-to-end, pinned here), attach-mode tunnel selection, startup lines never carrying the token.
- `main.parseArgs.test.ts` (+11): every validation rule and the refused combinations.
- `startServer.soapTunnelOrder.bun.test.ts`: boots the real `startServer` with a fake provider and a fake SOAP CSMS that calls back on BootNotification — the listener is bound when the provider runs, a restored charge point's first callback is served, and health / socket.io / MCP / `/` are 404 on the tunnel listener. Red under the previous ordering. `httpServer.soapCallbackOnly.test.ts` (11): the callback-only handlers. `startServer.soapTunnel.bun.test.ts`: `openSoapTunnel` wiring, fleet expansion. `lifecycle.test.ts`: exit code and teardown order.
- `CPRegistry.soapPublicBase.bun.test.ts` (7): derivation, explicit kept, refusal without base, not persisted / re-derived on restore, update re-derives, per-CP `soapPath`, WebSocket charge point untouched. `RegistryChargePointService.test.ts` (+1): clearing an explicit URL through the facade switches to derivation.
- `socketServer.rpc.test.ts` (+4): `server.info`, `cp.create` with/without a base. `httpServer.parseCreateBody.test.ts` (+1). `RemoteChargePointService.socket.test.ts` (+1).
- `ChargePointConfigModal.dom.test.tsx` (+4), `CpDetailPage.dom.test.tsx` (+4): optional field, preview, derived value not echoed back, Copy button and its failure state, tunnel note. `useServerInfo.dom.test.tsx` (3): shared fetch, refetch on reconnect, retry after failure.

Verified end to end with a real ngrok agent: spawn and attach (fleet of 2), callbacks routed through the tunnel to the right charge point, RPC-created charge point derived and reachable, `soap_callback_url` NULL in the state DB, no ngrok process left after SIGINT, agent death stops the daemon with code 1, post-tunnel startup failure kills the agent. After the callback-only listener: through the public origin `/v1/healthz`, `/socket.io/`, `/mcp` and `/` answer 404 while the callback route answers 405 (GET) / SOAP fault (unknown id), in both spawn and attach mode.

Pre-existing, unrelated: 28 vitest failures in `OCPPWebSocket.networkSim.test.ts` / `supervisionUrlFailover.test.ts` fail identically on `main`.

## Design note

The native `@ngrok/ngrok` SDK was considered and set aside: it would be the project's only native module (a risk for the `bun build --compile` sidecar), cannot attach to a sidecar agent, and its async close does not fit the daemon's synchronous shutdown path.

## Summary by CodeRabbit

* **New Features**
  * Added optional SOAP callback URL derivation from a configured public base URL or ngrok tunnel.
  * Added ngrok tunnel support with spawn and attach modes, configurable ports, domains, and API endpoints.
  * Tunnels now expose only the SOAP callback route; console, API, MCP, and Socket.IO remain local.
  * Added server information reporting for SOAP connectivity details.
  * Web console previews derived callback URLs and provides a Copy button.

* **Documentation**
  * Updated CLI, daemon, control-plane, transport, and web-console guidance for SOAP tunnels and derived URLs.


shiv3 (approved):
Thank you for this, @juherr — this is a really thorough PR. I especially like that the tunnel points at a SOAP-callback-only listener instead of the control plane, that the derived URL is deliberately not persisted, and that the docs, `--help`, flag tables and log entry all came in the same change. I also went through it locally (tsc clean, vitest 207 / bun 30 green on the head commit), so I'm approving.

Two things I noticed while reading, both small and fine as follow-ups (happy to send a PR for them myself if you'd rather):

1. **Restore without the tunnel.** After creating a SOAP charge point under `--soap-tunnel ngrok`, restarting the daemon *without* `--soap-tunnel` / `--soap-public-base-url` makes `restoreFromDatabase()` throw `OCPP SOAP versions require a callback URL …` from the `CLIChargePointService` constructor. There is no per-row try/catch, so the daemon exits 1 and none of the persisted charge points (including healthy WS ones) come back. Skipping that row with a warning that names the two flags would probably be enough, plus a test for "base gone on restore" (the current tests cover "base changed").
2. **MCP `server.info`.** `createRuntimeDeps(...)` in `startServer.ts` does not receive `serverInfo`, so `call_method server.info` over MCP returns the fallback while the Socket.IO path returns the real thing.

Re CodeRabbit's open thread about rolling back on startup failure: as far as I can tell it is already handled by design — `main()` exits 1 on a `startServer` rejection and the `process.once("exit")` hook in `soapTunnel.ts` kills the ngrok child — so I think that one can be resolved as is.

Thanks again for pushing #183 over the line.

juherr:
Thanks for the thorough review and for running it locally.

Both follow-ups are accurate — I checked them against the head commit:

1. `CPRegistry.restoreFromDatabase()` instantiates each row without a per-row guard, so a SOAP charge point created under `--soap-tunnel` takes the whole restore down when the daemon comes back without a base. Skipping that row with a warning naming `--soap-tunnel` / `--soap-public-base-url`, plus a "base gone on restore" test next to the "base changed" one, sounds right.
2. `createRuntimeDeps(...)` in `startServer.ts` indeed does not get `serverInfo`, so the MCP path answers the fallback. Passing it through (it's already computed for the Socket.IO deps) should be a one-liner plus a test.

Since you offered, I'm happy to leave both to you as a follow-up PR — thanks for taking them. And agreed on the CodeRabbit rollback thread: `main()` exits 1 on a `startServer` rejection and the `process.once("exit")` hook kills the ngrok child, so it can be resolved as is.

Thanks again for the approval.

shiv3:
Follow-up for the two items above is #356 (restore skips a SOAP charge point it cannot derive a callback for, row kept; MCP `server.info` gets the same `ServerInfo` as Socket.IO), with tests at both the registry and the daemon boundary. Resolved the three CodeRabbit threads as discussed: the startup-rollback one is addressed by design (`main()` exits 1 on a `startServer` rejection and the `process.once("exit")` hook kills the ngrok child), the other two were outdated.

Issue #314, opened 2026-09-04
Editing a blueprint, idTag or scenario file needs a restart to take effect


shiv3:
## Summary

Blueprints (#297) and idTag pools (#299) are both loadable from files. Editing one of those files today changes nothing until the charge point is recreated — the file is read once, at load, and the in-memory copy is authoritative from then on. For the human iterating on a blueprint by hand, that is a save → delete CP → recreate CP → reconnect loop for every one-line edit, and reconnecting throws away the CP's session state along with it.

Scenarios have the same shape: `src/utils/scenarios/` templates and the file a scenario is loaded from are snapshotted at load.

This is the last item on the fleet/load roadmap (`docs/analyses/fleet-load-and-observability-roadmap.md`, phase 6) and deliberately so — it only became meaningful once the things it would reload existed.

## Proposal

An opt-in `--watch` flag on the daemon that watches the files the process actually loaded — blueprint files, idTag files, scenario files — and re-reads one when it changes.

- `fs.watch` with a debounce. Editors write in bursts (write to temp, rename over), so an undebounced watch fires two or three times per save and, worse, can fire on a truncated intermediate file. Debounce and re-read; a parse failure keeps the previous good copy and logs, rather than leaving a CP with half a blueprint.
- **Opt-in, not default.** A daemon that silently re-reads files under it is surprising, and the agent-driven workflows this project is built around go through RPCs, where the control plane is already the source of truth. `--watch` is for the human editing a file.
- **An explicit mid-transaction rule.** A reload must not mutate a CP that is charging. New settings apply to the *next* session; an in-flight transaction runs to completion on the values it started with. Anything else makes a MeterValues stream change meaning halfway through, which is exactly the kind of thing that makes a trace unreadable.
- Emit a control-plane `event` on reload, so a connected console or agent can see that the file changed rather than inferring it from behaviour.

## Notes

- Reloading a **blueprint** affects CPs created from it *afterwards*, not retroactively — a blueprint is a template, and CPs created from it are independent. This should be stated in the docs, because the opposite is a plausible expectation.
- Reloading an **idTag pool** can safely apply live: the pool is drawn from per session, so the next draw picking up a new tag is coherent.
- Watching should degrade gracefully where `fs.watch` is unreliable (network mounts, some container filesystems): log once that watching is unavailable and carry on, rather than failing to start.

## Acceptance

[...]

PR #359, opened 2026-09-16
fix(control-plane): run_scenario_template keeps the auto-start gate out of its way and gains once: true (#318, #352)

### LINKED ISSUE #318: run_scenario_template fails with "already running" when the template auto-starts on load
## Summary

`run_scenario_template` crashes with `internal error` on the most ordinary input: a template whose start node is connect-triggered, run against a charge point that is already `Available`.

The shape is:

1. `runScenarioTemplate` loads the template.
2. `loadScenario` runs the auto-start gate. The charge point is connected, the template's start node triggers on connect, so **the scenario is already running** when the load returns.
3. `runScenarioTemplate` then issues its own explicit start.
4. That start throws "already running", and the whole RPC fails.

The caller sees a failure. The scenario is loaded and running anyway, as a side effect of the call that reported an error.

## How it was found

The identical defect in `run_scenario_file` was fixed in #314 (PR #317) — the explicit start there is now guarded by `isScenarioRunning`. `run_scenario_template` was observed to have the same shape at the same time and deliberately left alone: templates are not `--watch` sources, so it had no bearing on that issue, and changing an RPC's failure mode is a separate decision from the one #314 was making.

Filed so the observation is not lost.

## Notes

- The guard `run_scenario_file` now uses is the obvious candidate, but "make the explicit start idempotent" is a behaviour decision worth making deliberately: a caller currently cannot distinguish "your template was already running" from "your template started". Returning success either way is probably right — the post-condition the caller asked for holds — but that should be stated in `docs/concepts/control-plane.md` rather than inferred.
- Worth checking whether any other RPC does load-then-start against a gate that may already have started it.

## Acceptance

- `run_scenario_template` against an `Available` charge point with a connect-triggered template succeeds instead of returning `internal error`.
- A test covers it. Note the trap the #314 work hit: `cp.create` seeds "Essential CP Behavior", which auto-starts on connect and then owns the connector's auto-start slot, so a naive test never reaches the template's own start and passes with the bug present. The connector's seeded scenarios have to be cleared first.
- The success semantics for "already running" are documented.

### LINKED ISSUE #352: control-plane: a one-call way to run a template once (run_scenario_template once / load_scenario_template enabled:false)
## Summary

A headless caller has no one-call way to run a built-in template **once**. Since #253 a
`triggerOn: "connect"` scenario is armed per connection, so a completed template runs
again on the next reconnect — the right default for a long-lived simulator, and
documented in `docs/concepts/scenario-format.md` ("if a scenario must run only once per
process, drive it with an explicit `run_scenario`"). But every `cert16-*` template is
connect-triggered, and `run_scenario_template` is the only command that instantiates a
template: it loads the instance *enabled* and the auto-start walker owns it from there.
The documented run-once path therefore costs a caller **four** calls and a copy of the
definition, where one parameter would do.

## How it was found

`open-ocpp-tck` moved its pin from `0.7.5` to `0.7.9`
([juherr/open-ocpp-tck#136](https://github.com/juherr/open-ocpp-tck/pull/136)) and its
`TC_013` (`cert16-tc013-hard-reset`) went red on both CSMS drivers, confirmed in
isolation: after the `Reset(Hard)` the station reconnected and ran the template again —
a second `Authorize`, a second transaction — and the CSMS check read the newest
transaction, still open. On `0.7.5` the same trace ends at the post-reboot
`BootNotification`.

## What the caller does today

The sequence in
[`tck/template-once.ts`](https://github.com/juherr/open-ocpp-tck/blob/main/tck/template-once.ts),
all over JSON Lines with ids, **before `connect`** so the walker cannot fire at load:

1. `load_scenario_template {connector, templateId}` → `{scenarioId}`
2. `get_scenario {connector, scenarioId}` → the instance the template minted
3. `remove_scenario {connector, scenarioId}`
4. `load_scenario {connector, scenario: {...definition, enabled: false}}`

then, after the boot gate, `run_scenario {connector, scenarioId}`.

It works because `tryAutoStartForConnector` skips `def.enabled === false`
(`src/cli/service.ts:2366`, `:2455`) while `runScenario` (`:1201`) does not consult
`enabled`. What makes it four calls is that a template's instance only exists after
`loadScenarioTemplate`, and `loadScenario` kicks the walker on an Available station
(`:1013`) — so the enabled instance must never meet a connected station, and the caller
has to place the whole sequence before `connect`.

## Ask

One of, in order of preference:

1. **`run_scenario_template` gains `once: true`** (or `autoStart: false`): load the
   instance with `enabled: false`, then `runScenario` it. One call, and for that caller
   it also closes #318 by construction — a disabled instance cannot have auto-started
   before the explicit start, so there is no `already running` to refuse.
2. **`load_scenario_template` gains `enabled: false`**, passed through to the instance.
   Two calls (`load_scenario_template` + `run_scenario`), no definition copy.
3. **`get_scenario_template {templateId, connector}`** returning the instance
   `createScenario` would mint, without loading it. Still two calls, but the caller
   owns the definition it then `load_scenario`s.

Whichever lands, a line in `docs/concepts/control-plane.md` beside `run_scenario_template`
saying which of "run once" / "arm on every connect" the caller gets by default.

## Why upstream rather than the caller

The four-call sequence is fine to keep, and the caller pins it with a guard against a
model of the walker — but that model is the guard's one assumption, and #346's item 1
(the JSON-Lines surface as a stable contract) is where a one-parameter form belongs: a
caller reshaping a definition it read back is coupled to the instance shape
`createScenario` produces, where `once: true` is coupled to nothing.

## References

- #253 — the re-arm, and its "explicit `run_scenario`" advice
- #318 — `run_scenario_template` refused with `already running` when the load auto-starts
- #346 — the stable contract towards `open-ocpp-tck`, item 1
- juherr/open-ocpp-tck#136 — the pin move and the sequence above

shiv3:
Closes #318. Closes #352 (option 1 of the three the issue lists, which it prefers and which closes #318 by construction).

## #318 — "already running" on the most ordinary input

Every `cert16-*` template starts on `triggerOn: "connect"`. Loading one onto an Available charge point ran the auto-start gate, so the instance was already running when `run_scenario_template`'s own explicit start ran, and the RPC failed with `internal error` for a scenario it had in fact loaded and started. The load now passes `{ autoStart: false }` — the shape `run_scenario_file` has had since #314 (PR #317), per the issue's follow-up comment, and not the `isScenarioRunning` guard that was tried and reverted there — and the explicit start is the one run the call makes, carrying the caller's `strict`.

## #352 — `run_scenario_template { "once": true }`

Loads the template's instance with `enabled: false` before that start, so the connect-triggered walker never re-arms it on a later reconnect (`tryAutoStartForConnector` skips disabled definitions; `runScenario` does not consult `enabled`). Without it the instance stays enabled and re-arms on every connect, as every loaded scenario does since #253 — the docs row now says so, as the issue asked. This collapses open-ocpp-tck's four-call `tck/template-once.ts` sequence to one call.

## Surfaces (#316 rule: all together)

`src/protocol/methods.ts` (`once?: boolean`), Socket.IO dispatch, MCP `run_scenario_template` tool (`mcpToolSchemaParity` green), JSON-Lines (`once` and `strict` — this surface silently dropped `strict` before), `ScenarioRunOptions.once`, `RegistryChargePointService`, `RemoteChargePointService` (now also forwards `strict`), `LocalChargePointService` (browser), and the standalone single-CP target. Not added: #352's options 2 (`load_scenario_template { enabled: false }`) and 3 (`get_scenario_template`).

## Tests

- `src/cli/server/__tests__/runScenarioTemplate.once.bun.test.ts` — through `createRuntimeDeps().chargePointService` against a real `CLIChargePointService` set to Available, with the seeded default cleared first (the trap the issue names). A control row proves the default load path does auto-start the template in this harness; then: the RPC succeeds and the run is the one it started; without `once` the instance re-arms on the next connect; with `once` it is loaded disabled and does not. 3 of 4 fail against the old `RegistryChargePointService` (`Scenario … is already running`).
- `src/cli/__tests__/jsonMode.runScenarioTemplate.test.ts` — `once` / `strict` forwarded, omitted when absent, non-boolean rejected.
- `socketServer.rpc.test.ts` expectation gains the `{ autoStart: false }` argument.

Locally: tsc clean, eslint clean, bun 915/915, targeted vitest green. The full vitest run had two files time out at 5 s (`useChargePoints.registry`, `DataProvider.detect`) and `main.parseArgs` timed out in bun — both reproduce identically on an unchanged `main` checkout under the same load (an unrelated build was holding the machine at a load average above 200), so CI is the authoritative run here.

## Wiki (CLAUDE.md ingest)

[Control plane → CP command methods](docs/concepts/control-plane.md#cp-command-methods): `run_scenario_template` row (params, gate behaviour, which of run-once / re-arm is the default). [Scenario format](docs/concepts/scenario-format.md): run-once advice names `once: true`; the `strict` note covers both RPCs. [MCP endpoint](docs/entities/mcp-endpoint.md) tool row. [GitHub issues](docs/sources/github-issues.md): #318 and #352 rows. `log.md` entry, `updated:` bumped, link check 0.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `once` option to `run_scenario_template`.
  - When enabled, the scenario runs once and does not restart on later reconnects.
  - By default, scenarios remain enabled and re-arm on every connection.
  - Strict execution settings are now honored consistently across local, remote, JSON, and MCP interfaces.

- **Bug Fixes**
  - Prevented connect-triggered scenarios from running prematurely and causing “already running” failures.

- **Documentation**
  - Updated scenario, control-plane, MCP endpoint, and issue reference documentation.


shiv3:
Re-ran the three load-sensitive files once the unrelated build finished (load average back to ~6): `useChargePoints.registry.test.ts` and `DataProvider.detect.test.tsx` 3/3, `bun test src/cli/__tests__` 200/200. So the local record is now fully green too.

